PR #6, opened 2017-11-22
Add *.g8s wildcard

r7vme:
In Azure we need to have explicit *.g8s for
services like happa, api, etc. to work.

We don't use ingress dns at all, so just remove it.

This is not required in AWS.

### DIFF
--- a/modules/azure/vnet/lb-ingress.tf
+++ b/modules/azure/vnet/lb-ingress.tf
@@ -25,8 +25,11 @@ resource "azurerm_public_ip" "ingress_ip" {
   }
 }
 
-resource "azurerm_dns_a_record" "ingress_dns" {
-  name                = "${var.ingress_dns}"
+# We need explicitly add this entry for apps like happa.g8s.<base> to be working.
+# In Azure this is required (not needed in AWS) and looks like it's aligned with RFC.
+# https://tools.ietf.org/html/rfc1034#section-4.3.3
+resource "azurerm_dns_a_record" "g8s_wildcard_dns" {
+  name                = "*.${var.api_dns}"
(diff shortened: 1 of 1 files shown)


kopiczko:
Why we don't use ingress for happa, api, etc.?

r7vme:
We use ingress, we just need explicit wildcard DNS for ingress.

`*.g8s.basedomain --> ingress load balancer`

For AWS we don't need that and wildcard like this `*.basedomain --> ingress load balancer` is enough.

kopiczko:
> we just need explicit wildcard DNS for ingress

Does it mean that we'll have to create DNS entry for every guest cluster ingress our customer create?

r7vme:
no no. All will be working fine for any dns names as before.

Example, if you have
```
*.domain -> IP1
b.domain -> IP2
*.b.domain -> IP1
```

And want to resolve `a.b.domain`, so it will be reolved to IP1. Problem is that AWS allows to resolve this w/o 3rd entry (*.b.domain --> IP1), that's why we did not have 3rd entry in AWS. Azure just drops `a.b.domain` if there is b.domain, but no *.b.domain

As per [RFC](https://tools.ietf.org/html/rfc1034#section-4.3.3) this looks right that Azure drops this :)

>   - When the query name or a name between the wildcard domain and
     the query name is know to exist.  For example, if a wildcard
     RR has an owner name of "*.X", and the zone also contains RRs
     attached to B.X, the wildcards would apply to queries for name
     Z.X (presuming there is no explicit information for Z.X), but
     not to B.X, A.B.X, or X.

kopiczko:
I think we have a wildcard on AWS as well.

r7vme:
we do, but only one.

PR #7, opened 2017-11-22
Docs. Restrict service principal permissions.

r7vme:
1) We are not going to use MSI (own vm identity) in near future as it's
brokne in 1.7 and 1.8 (only 1.9 is OK). So we need to bound permissions
for service principal. Allow operations in cluster resource group only.

2) Add note to upload secure credentials into keepass

3) Clean up service principal if cluster deleted

Related to: https://github.com/giantswarm/azure-terraform/issues/5

### DIFF
--- a/docs/installation-guide-azure.md
+++ b/docs/installation-guide-azure.md
@@ -38,13 +38,20 @@ az storage account keys list -g ${NAME}-terraform  --account-name ${NAME}terrafo
 
 ### Create service principal
 
-To get subscription use `az account list | jq '.[].id'`
+Create resource group for cluster. We need one to assign permissions.
 
 ```
-az ad sp create-for-rbac --name=${NAME}-sp --role="Contributor" --scopes="/subscriptions/XXX..."
+az group create -n ${NAME} -l westeurope
 ```
 
-Make sure to save output, it will be needed in next step.
+Create service principal with permissions limited to resource group.
+
+```
+export SUBSCRIPTION_ID=$(az account list | jq '.[].id')
+az ad sp create-for-rbac --name=${NAME}-sp --role="Contributor" --scopes="/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${NAME}"
+```
+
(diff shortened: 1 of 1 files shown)


docs/installation-guide-azure.md
```diff
@@ -38,13 +38,20 @@ az storage account keys list -g ${NAME}-terraform  --account-name ${NAME}terrafo
...
 
-Make sure to save output, it will be needed in next step.
+Create service principal with permissions limited to resource group.
+
+```
+export SUBSCRIPTION_ID=$(az account list | jq '.[].id')
+az ad sp create-for-rbac --name=${NAME}-sp --role="Contributor" --scopes="/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${NAME}"
+```
+
+Please save these and storage credentials above in keepass (e.g. "<cluster name> azure host cluster credentials"). They will be needed in next step.
```
kopiczko: Why do we need that? What uses the credentials?

r7vme: 1) Storage account used to store terraform state in Azure storage (See envs.sh)
2) Service Principal used by k8s Azure cloud provider to access Azure API (in the future we will use MSI, but it's not the case in near future)

kopiczko: Do we need to use the cloud provider now?

r7vme: What do you mean use cloud provider?

We are using cloud provider and we need to configure it (See [here](https://github.com/giantswarm/azure-terraform/blob/master/cloud-config/master.yaml.tmpl#L902-L916)). Only way to avoid this to use MSI, which is not the case for <1.9

JFYI: In AWS MSI-like stuff works "out-of-the-box" we just need to assign IAM policy to VM, so in AWS we don't need any AWS credentials for k8s :stuck_out_tongue_closed_eyes:

kopiczko: Yes, we need that to create LB I guess.
Let's also create an follow up issue to make sure we change that.

r7vme: yep, follow up here https://github.com/giantswarm/azure-terraform/issues/5

PR #9, opened 2017-12-19
Use managed disks as default storage class

r7vme:
Switch default storage class to managed disks.

By default Azure creates blob and tries to attach it to instance, this works only for VMs with unmanaged disks (Old-style), we use managed disks VMs.


r7vme:
Indeed :)

PR #13, opened 2017-12-29
Upgrade Calico to 3.0.1

r7vme:
Not tested yet.

### DIFF
--- a/cloud-config/master.yaml.tmpl
+++ b/cloud-config/master.yaml.tmpl
@@ -117,17 +117,18 @@ write_files:
   owner: root
   permissions: 644
   content: |
-    # Source: https://docs.projectcalico.org/v2.6/getting-started/kubernetes/installation/hosted/kubernetes-datastore/policy-only/1.7/calico.yaml
+    #
+    # Source: https://docs.projectcalico.org/v3.0/getting-started/kubernetes/installation/hosted/kubernetes-datastore/policy-only/1.7/calico.yaml
     #
     # Changed values:
     #   - CALICO_IPV4POOL_IPIP = off
     #   - CALICO_IPV4POOL_CIDR = parameterized
     #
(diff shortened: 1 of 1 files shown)


cloud-config/master.yaml.tmpl
```diff
@@ -381,12 +520,27 @@ write_files:
...
+      scope: Namespaced
+      group: crd.projectcalico.org
+      version: v1
+      names:
+        kind: NetworkPolicy
+        plural: networkpolicies
+        singular: networkpolicy
+
+    ---
+
```
kopiczko: Is this the storage?

cloud-config/master.yaml.tmpl
```diff
@@ -136,32 +137,145 @@ write_files:
...
+              periodSeconds: 30
+              initialDelaySeconds: 30
+            readinessProbe:
+              httpGet:
+                path: /readiness
+                port: 9098
+              periodSeconds: 10
+
+    ---
+
```
kopiczko: Do we use Typha in guest clusters?

r7vme: We still use old calico in Azure guest clusters and need to update manifests too. So Typha will be added.

cloud-config/master.yaml.tmpl
r7vme: Exactly

r7vme:
Applied to gollum.

Additional migration steps are done manually.
1) Delete all Calico CRDs
2) Delete new Calico with `kubectl delete -f /srv/calico-all.yaml`
3) Apply back `kubectl apply -f /srv/calico-all.yaml`

These steps only for Azure (k8s datastore) for AWS and baremetal they probably not needed, as new Calico by default will use etcd v3 (which is empty for Calico)

PR #14, opened 2017-12-29
Use local state for cloud configs

r7vme:
We don't really need any state for cloud-configs generation.


r7vme:
State is not needed, because we use hack to generate and gzip cloud-configs. State does not reflect contents of these files, only files existence. So state can not say if smth changed or not.

PR #19, opened 2018-02-01
Remove node-exporter in favor kubernetes managed one

r7vme:
we already have node exporter managed by g8s-kubernetes


JosephSalisbury:
i should have cleaned this up (and aws too), thanks for taking care <3

PR #21, opened 2018-02-01
Fix etcd defragmentation service. Use 127.0.0.1 for etcd

r7vme:
In azure we can not use load balancer if it points to the same node.


kopiczko:
This is BS. I can't see how ppl are not complaining about that.

PR #22, opened 2018-02-03
Replace kube-dns with coredns

corest:
Towards: https://github.com/giantswarm/giantswarm/issues/2135

Requires RBAC changes to be merged for further update

### DIFF
--- a/cloud-config/master.yaml.tmpl
+++ b/cloud-config/master.yaml.tmpl
@@ -588,24 +588,22 @@ write_files:
     parameters:
       kind: Managed
       storageaccounttype: Standard_LRS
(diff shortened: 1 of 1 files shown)


r7vme:
can you please take care to clean up kube-dns before rolling out terraform

i think we need to execute similar commands https://github.com/giantswarm/k8scloudconfig/blob/master/v_2_0_0/master_template.go#L1403-L1407

corest:
Sure
I forgot that we have here auto-update )

r7vme:
workflow can be following:
1) merge
2) job will be triggered (max in 1 hour after merge)
3) clean up kube-dns
4) approve job

PR #23, opened 2018-02-07
Calico fix RBAC rules

corest:
As we are using calico with kubernetes datastore, `calico-node` cluster role requires different rules from those we are using with etcd backed calico. Rules are taken from upstream: https://github.com/projectcalico/calico/blob/master/v3.0/getting-started/kubernetes/installation/hosted/rbac-kdd.yaml
Also this PR removes calico policy-controller related RBAC stuff

### DIFF
--- a/cloud-config/master.yaml.tmpl
+++ b/cloud-config/master.yaml.tmpl
@@ -965,10 +965,66 @@ write_files:
     rules:
       - apiGroups: [""]
         resources:
+          - namespaces
+        verbs:
+          - get
+          - list
+          - watch
+      - apiGroups: [""]
+        resources:
+          - pods/status
+        verbs:
+          - update
+      - apiGroups: [""]
+        resources:
           - pods
+        verbs:
+          - get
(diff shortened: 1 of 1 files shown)


calvix:
why are we removing calico policy controller?

corest:
I'm removing rbac related stuff as there is no policy-controller in gollum. Why? @r7vme :)

r7vme:
calico uses k8s datastore in Azure. So calico-node can grab policies directly from k8s, so no need in additional transtator (from k8s to etcd).

calvix:
I thought calico-policy-controller is doing the network firewall magic, or is this already included in calico-kube-controllers ?

r7vme:
no controllers at all. Only calico-node (felix).

r7vme:
and only for policies. as routing and ipam done by k8s (+azure routing tables)

calvix:
My thinking was that policies are managed by calico-policy-controller

PR #25, opened 2018-02-09
giantnetes modules for aws installations

r7vme:
This is first drop with aws terraform modules.

Changes:
- switched to ignition
- using similar to Azure module structure

Tasks:
- [x] common modules
- [x] bastion
- [x] vault
- [x] master
- [x] workers
- [x] vpn

NOTE: this repo will be renamed to `giantnetes-terraform`.


calvix:
Looks so far good. I really like the user ignition separation. 👍

modules/aws/bastion/bastion.tf
```diff
@@ -0,0 +1,76 @@
...
+  source_dest_check           = false
+  subnet_id                   = "${var.bastion_subnet_ids[count.index]}"
+  vpc_security_group_ids      = ["${aws_security_group.bastion.id}"]
+
+  root_block_device = {
+    volume_type = "${var.volume_type}"
+    volume_size = "${var.root_volume_size}"
+  }
+
+  user_data = "${var.user_data}"
```
r7vme: do not use S3 for user data for bastion.

r7vme: I will use S3 for master and workers only as they have 16kb+ user_data

modules/aws/vpc/vpc.tf
```diff
@@ -0,0 +1,144 @@
+# Following networking schema implemented:
```
r7vme: copy/paste of one of the Timo's git comments, which is really useful

platforms/aws/giantnetes/backend.tf
```diff
@@ -0,0 +1 @@
+../../../build/backend.tf
```
r7vme: same as for Azure. Symlink will point to cluster-specific tf state backend configuration

platforms/aws/giantnetes/main.tf
```diff
@@ -0,0 +1,91 @@
+module "container_linux" {
```
r7vme: this module just returns version number for "latest" coreos.

platforms/aws/giantnetes/main.tf
```diff
@@ -0,0 +1,91 @@
...
+  source = "../../../modules/container-linux"
+
+  coreos_channel = "${var.container_linux_channel}"
+  coreos_version = "${var.container_linux_version}"
+}
+
+# Get ami ID for specific Container Linux version.
```
r7vme: find out ami ID

platforms/aws/giantnetes/main.tf
```diff
@@ -0,0 +1,91 @@
...
+# Create S3 bucket for ignition configs.
+module "s3" {
+  source = "../../../modules/aws/s3"
+
+  aws_account  = "${var.aws_account}"
+  cluster_name = "${var.cluster_name}"
+}
+
+locals {
+  ignition_users = "${file("${path.module}/../../../ignition/users.yaml")}"
```
r7vme: Idea is to use same users (`passwd`) block for all resources.
1) avoid copy/pasting
2) we can generate users.yaml on the fly (e.g. with `opsctl`)

platforms/aws/giantnetes/main.tf
```diff
@@ -0,0 +1,91 @@
...
+}
+
+# Generate ignition config for bastions.
```
r7vme: bastion ignition template

platforms/aws/giantnetes/main.tf
```diff
@@ -0,0 +1,91 @@
...
+data "template_file" "bastion" {
+  template = "${file("${path.module}/../../../ignition/bastion.yaml.tmpl")}"
+}
+
+# Convert ignition config to raw json.
+data "ct_config" "bastion" {
```
r7vme: here is the interesting part. I've looked on possibility to use [ignition_config](https://www.terraform.io/docs/providers/ignition/d/config.html), but it does not fit well for our case. Problem is that i need to define every particular part of ignition template in terraform. e.g. how to define [systemd unit](https://www.terraform.io/docs/providers/ignition/d/systemd_unit.html#example-usage). I don't like it.

So i took alternative approach.
1) generate ignition from yaml-based "container linux configuration" files (similar to cloud-config)
2) convert to raw json ignition with `ct` provider. (`ct` is official coreos tool to convert from yaml to ignition)

One important thing: [ct terraform provider](https://github.com/coreos/terraform-provider-ct/issues/21) is still not in official terraform repos. Latest release 0.2.0 had issues with user "groups", but latest dev version worked fine for me.

calvix:
nice progress

ignition/aws/master.yaml.tmpl
```diff
@@ -0,0 +1,2070 @@
...
+          apiVersion: v1
+          kind: ServiceAccount
+          metadata:
+            name: calico-kube-controllers
+            namespace: kube-system
+    - path: /srv/calico-node-sa.yaml
+      filesystem: root
+      mode: 0644
+      contents:
+        inline: |
```
calvix: interesting, i have not seen `inline` in the default ignition 
i guess its part of `ct`

r7vme: i'm using ct. No other way to manage "raw" json ignition. 

Look here https://github.com/giantswarm/giantnetes-terraform/blob/dev/platforms/aws/giantnetes/main.tf#L71

Unfortunately we need to get ct_config provider manually, because it's not in official TF repos. :(

calvix: that sad, but atleast its working
and when we will have CD we dont have to care that much as it  will be part of the CD config

calvix:
wanna whole review now ?

r7vme:
i think so, it's already hard, because so many changes.

few things are still missing:
- we need gen-provider.sh script to generate customer spcific provider.tf file. Right now only working for giantswarm account
- migration path for existing clusters.

r7vme:
gentle ask for review/approve. `gauss` migrated to this code.

docs/installation-guide-aws.md
```diff
@@ -0,0 +1,147 @@
...
+
+NOTE: Reexecute `source envs.sh` in every new console.
+
+## Install
+
+Install consists two stages:
+- Vault (only needed because we bootstrapping Vault manually)
+- Kubernetes
+
+Master and workers will be created with in the Vault stage and expectedly will fail (and recreated later). This is done to keep single Terraform state and simplify cluster management after installation. Master and workers will be reprovisioned with right configuration in the second state called Kubernetes.
```
corest: with in -> within
Why do we need to create master/workers within vault stage?

docs/installation-guide-aws.md
```diff
@@ -0,0 +1,147 @@
...
+Set "TF_VAR_nodes_vault_token" in envs.sh with node token that was outputed by Ansible.
+
+### Stage: Kubernetes
+
+```
+source envs.sh
+```
+
+```
+terraform init ../platforms/azure/giantnetes
```
corest: This confuses me a bit. Why stage Vault is bounded to `platforms/aws` and stage Kubernetes - to `platforms/azure`? Such example denies idea of single state management

docs/installation-guide-aws.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Prerequisites
+
+Common:
+- `aws` cli
+
+### Create S3 bucket and DynamoDB table for terraform state
+
+```
+export CLUSTER="cluster1"
```
calvix: IN few lines below there is used 'gauss' instead of 'cluster1'  I would pick one and stay consistent

docs/installation-guide-aws.md
r7vme: because we have one terraform manifest for whole cluster and Vault still provisioned with Ansible.

docs/installation-guide-aws.md
r7vme: typo :)

r7vme:
thanks for comments all addressed.

PR #31, opened 2018-02-22
Add support for aws profiles

r7vme:
this allows to use AWS profiles. Including case with assume role.

Only limitation kind of we need to agree that we will define profiles with codename of the clusters. Not with the name of the customer as we are doing at the moment.


r7vme:
ping @calvix @corest

PR #36, opened 2018-02-23
Fix typos in manifest naming for azure

r7vme:
we found that these two manifests were not applied for a while.

### DIFF
--- a/cloud-config/master.yaml.tmpl
+++ b/cloud-config/master.yaml.tmpl
@@ -784,7 +784,7 @@ write_files:
       variables-hash-bucket-size: "128"
       server-name-hash-bucket-size: "1024"
       server-name-hash-max-size: "1024"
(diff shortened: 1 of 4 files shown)


cloud-config/master.yaml.tmpl
```diff
@@ -784,7 +784,7 @@ write_files:
       variables-hash-bucket-size: "128"
       server-name-hash-bucket-size: "1024"
       server-name-hash-max-size: "1024"
-- path: /srv/ingress-controller-sa.yaml
+- path: /srv/ingress-controller-sa.yml
```
calvix: I would rather make the changes towards `yaml` as we use that for most of our files in and in  other projects too.

r7vme: Makes sense. See new changes

PR #38, opened 2018-02-26
DOCS: Fix typos in azure guide.

r7vme:
just some boyscouting.

### DIFF
--- a/docs/installation-guide-azure.md
+++ b/docs/installation-guide-azure.md
@@ -8,7 +8,7 @@ Common:
 ### Create storage account for terraform state
 
 ```
-export NAME="cluster5"
+export NAME="cluster1"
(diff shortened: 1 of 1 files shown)


docs/installation-guide-azure.md
```diff
@@ -198,15 +198,22 @@ cd build
...
+export NAME=cluster1
 for i in envs.sh backend.tf; do
   az storage blob download --account-name ${NAME}terraform -c ${NAME}-build -n ${i} -f ${i}
 done
 ```
 
+Command below will ask for secrets that can be found in keepass.
+
+```
+source envs.sh
```
corest: So, is that one for all, or creds per account?

r7vme: per account

PR #43, opened 2018-03-02
Use protocol 4 for IPIP

r7vme:
In calico slack they recommended to use protocol 4 instead of 94 in security groups.

Unfortunately i can not confirm that it fixes the IPIP tunnels problem, but at least this does not break anything. All last roll outs done with this change.

### DIFF
--- a/modules/aws/master/master.tf
+++ b/modules/aws/master/master.tf
@@ -97,7 +97,7 @@ resource "aws_security_group" "master" {
   ingress {
     from_port   = 0
     to_port     = 0
-    protocol    = 94
+    protocol    = 4
     cidr_blocks = ["${var.vpc_cidr}"]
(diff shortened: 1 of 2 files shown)


modules/aws/master/master.tf
```diff
@@ -97,7 +97,7 @@ resource "aws_security_group" "master" {
   ingress {
     from_port   = 0
     to_port     = 0
-    protocol    = 94
+    protocol    = 4
```
calvix: what does this mean? any link to docs or something like that? or is it just some magic value only AWS knows ?

r7vme: here we go https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml

PR #44, opened 2018-03-05
Restrict bastion ssh rule for VPN case.

r7vme:
Use IPsec subnet for ssh access if using VPN.

### DIFF
--- a/modules/aws/bastion/bastion.tf
+++ b/modules/aws/bastion/bastion.tf
@@ -1,3 +1,10 @@
+locals {
+  default_ssh_access_subnet = "0.0.0.0/0"
+
+  # If behind VPN allow SSH access only from VPN subnet.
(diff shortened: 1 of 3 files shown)


calvix:
where do we configure this ? in envs.sh ?

r7vme:
Right, it's in `envs.sh` see sample https://github.com/giantswarm/giantnetes-terraform/blob/master/examples/aws/example-build/envs.sh#L17

r7vme:
> how we will track updates in envs.sh? Separate issue in SRE board?

JFYI `envs.sh` was not changed for this change. We are storing `envs.sh` in S3 versioned bucket. In general, it's a good idea for transparency to store it in git, we can do that. Created issue https://github.com/giantswarm/giantswarm/issues/2733

PR #48, opened 2018-03-15
Use new tagging for AWS

r7vme:
New tags:
- giantswarm.io/installation=CLUSTER_NAME
- kubernetes.io/cluster/CLUSTER_NAME=owned

Instead of:
- Environment=CLUSTER_NAME
- KubernetesCluster=CLUSTER_NAME

Terraform does not support interpolation for tag name that's why such ugly constructions are used. See bug https://github.com/hashicorp/terraform/issues/14516

### DIFF
--- a/modules/aws/master/master.tf
+++ b/modules/aws/master/master.tf
@@ -1,3 +1,10 @@
+locals {
+  common_tags = "${map(
+    "giantswarm.io/installation", "${var.cluster_name}",
+    "kubernetes.io/cluster/${var.cluster_name}", "owned"
+  )}"
+}
+
 data "aws_availability_zones" "available" {}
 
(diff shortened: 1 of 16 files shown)


modules/aws/master/master.tf
```diff
@@ -18,22 +25,25 @@ resource "aws_instance" "master" {
 
   user_data = "${data.ignition_config.s3.rendered}"
 
-  tags = {
-    Name              = "${var.cluster_name}-master${count.index}"
-    Environment       = "${var.cluster_name}"
-    KubernetesCluster = "${var.cluster_name}"
-  }
+  tags = "${merge(
```
corest: Using merge is nice

PR #55, opened 2018-03-28
Add SSH key for Tuomas to cloud-config templates

tuommaki:


calvix:
this now needs to be reinstalled

PR #61, opened 2018-04-04
Enable S3 access logs for existing buckets

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/2726


pipo02mix:
I will wait to create the access log in all AWS installations before merging, because I suspect it will fail if the bucket does not exist and conveyor try to roll it out

pipo02mix:
ok now all host clusters have delivery log bucket created and tf state bucket forward logs to it. So after tf roll out changes ignition bucket should also have it

teemow:
wouldn't it make sense to add the bucket creation itself to terraform too?

PR #68, opened 2018-04-12
Add e2e conformance testing for AWS

r7vme:


misc/e2e-aws.sh
```diff
@@ -22,7 +22,7 @@ WORKDIR=$(pwd)
 BUILDDIR=${WORKDIR}/build
 CLUSTER=e2e-cluster-$(echo ${CIRCLE_SHA1} | cut -c 1-6)
 SSH_USER="e2e"
-KUBECTL_CMD="docker run --net=host --rm quay.io/giantswarm/docker-kubectl:f51f93c30d27927d2b33122994c0929b3e6f2432"
+KUBECTL_CMD="docker run -i --net=host --rm quay.io/giantswarm/docker-kubectl:f51f93c30d27927d2b33122994c0929b3e6f2432"
```
r7vme: we need that to be able to read `stdin` for `kubectl apply`. See below

misc/e2e-conformance-pod.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: e2e
+spec:
+  containers:
+  - name: e2e
+    image: gcr.io/heptio-images/kube-conformance:v1.9
```
r7vme: JFYI: when we will do upgrade to new version we also need to bump it here.

misc/e2e-aws.sh
```diff
@@ -226,7 +226,7 @@ stage-destroy() {
   terraform init ../platforms/aws/giantnetes
   terraform destroy -force ../platforms/aws/giantnetes
 
-  aws s3 rb s3://${CLUSTER}-access-logs
+  aws s3 rb s3://${CLUSTER}-access-logs --force
```
r7vme: `--force` allows to delete bucket with objects.

misc/e2e-conformance-pod.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: e2e
+spec:
+  containers:
+  - name: e2e
+    image: gcr.io/heptio-images/kube-conformance:v1.10
```
r7vme: JFYI: we also need to update version, when updating kubernetes

r7vme:
gentle ask for review @calvix @corest @fgimenez

fgimenez:
@r7vme ok thx. I was thinking mainly of reusing the code here for running the conformance suite, but you are right this is not the better place. We could use https://raw.githubusercontent.com/giantswarm/giantnetes-terraform/master/misc/e2e-conformance-pod.yaml from jobs in other repos, is that correct?

r7vme:
Sure, i would just copy/paste it to your repo to not depend on this repo, as host cluster and guest cluster versions can vary.

PR #70, opened 2018-04-12
Add e2e CI check for Azure

r7vme:
adds CI build script that capable to build cluster and checks for `Ready` nodes.

conformance tests coming.

### DIFF
--- a/misc/e2e-azure.sh
+++ b/misc/e2e-azure.sh
@@ -1,0 +1,285 @@
+#!/bin/bash
+
+# Copyright 2018 Giant Swarm.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


misc/e2e-azure.sh
```diff
@@ -0,0 +1,285 @@
...
+EOF
+
+    # Bootstrap insecure Vault.
+    export ANSIBLE_HOST_KEY_CHECKING=False
+    ansible-playbook -i hosts_inventory/${CLUSTER} -e dc=${CLUSTER} bootstrap1.yml
+
+    # Init Vault with one unencrypted unseal key.
+    # NOTE: sed here is to filter ANSI escape codes and color codes.
+    init_output=$(exec_on vault1 vault operator init -key-shares=1 -key-threshold=1 |\
+        sed -r -e 's/\x1B\[([0-9]{1,2}(;[0-9]{1,2})?)?[m|K]//g' -e 's/[\x01-\x1F\x7F]//g')
```
corest: :open_mouth:

r7vme:
We don't have any peering for host cluster (only guest cluster creates peering), networking completely isolated.

misc/e2e-azure.sh
r7vme: did not find better way unfortunately :/

PR #71, opened 2018-04-12
Migrate azure to ignition

calvix:
migrate azure to ignition - start with bastion,vault and worker

Update added also master.

Few notes:
* we use same ignition for bastion and vault  for aws/azure
* all users are now in one file only  for both aws and azure
* master ignition is too big to be passed as user data so we pass only super simple ignition which has info where it should fetch the real big ignition
* the big ignition is uploaded to Azure blob storage
* added boot diagnostic to master instance


calvix:
Should be final, this now works nicely, I tried recreating from freash few times and all seem to be good.

modules/azure/master-as/variables.tf
```diff
@@ -66,3 +66,8 @@ variable "api_backend_address_pool_id" {
   type        = "string"
   description = "API load balances address pool id."
 }
+
+variable "boot_diagnostics_storage_uri" {
+  type        = "string"
+  description = "storage account uri fro boot diagnostics"
```
r7vme: nitpick. Please start from upper case and there is typo in `fro`

platforms/azure/giantnetes/main.tf
```diff
@@ -81,14 +100,89 @@ module "vault" {
...
+  platform     = "azure"
+  pretty_print = false
+}
+
+resource "local_file" "master_ignition_big" {
+  content  = "${data.ct_config.master_big.rendered}"
+  filename = "${path.cwd}/generated/master-ignition.yaml"
+}
+
+resource "azurerm_storage_account" "storage_acc" {
```
r7vme: can you please move all storage related part to separate module. Like it's done for S3 buckets

https://github.com/giantswarm/giantnetes-terraform/tree/master/modules/aws/s3

platforms/azure/giantnetes/main.tf
```diff
@@ -81,14 +100,89 @@ module "vault" {
...
+}
+
+resource "azurerm_storage_container" "ignition" {
+  name                  = "ignition"
+  resource_group_name   = "${module.resource_group.name}"
+  storage_account_name  = "${azurerm_storage_account.storage_acc.name}"
+  container_access_type = "container"
+}
+
+resource "azurerm_storage_blob" "ignition_blob" {
```
r7vme: please move ignition_blob and loader to master module. Like done for aws 

https://github.com/giantswarm/giantnetes-terraform/blob/master/modules/aws/master/master.tf#L142-L165

platforms/azure/giantnetes/variables.tf
```diff
@@ -67,7 +67,9 @@ variable "vault_storage_type" {
 }
 
 variable "master_vm_size" {
-  type    = "string"
+  type = "string"
+
+  #default = "Standard_D2s_v3"
```
r7vme: Let's switch to v3 in separate PR

platforms/azure/giantnetes/main.tf
```diff
@@ -81,14 +100,89 @@ module "vault" {
...
+    "DOCKER_CIDR"              = "${var.docker_cidr}"
+    "ETCD_DOMAIN_NAME"         = "${var.etcd_dns}.${var.base_domain}"
+    "G8S_VAULT_TOKEN"          = "${var.nodes_vault_token}"
+    "K8S_DNS_IP"               = "${var.k8s_dns_ip}"
+    "VAULT_DOMAIN_NAME"        = "${var.vault_dns}.${var.base_domain}"
+  }
+}
+
+# Convert ignition config to raw json and merge users part.
+data "ct_config" "master_big" {
```
r7vme: may be remove `_big` suffix as you switched to `ignition_config` already.

platforms/azure/giantnetes/main.tf
```diff
@@ -81,14 +100,89 @@ module "vault" {
...
+  content      = "${format("%s\n%s", local.ignition_users, data.template_file.master_big.rendered)}"
+  platform     = "azure"
+  pretty_print = false
+}
+
+resource "local_file" "master_ignition_big" {
```
r7vme: let's also move this to master module

platforms/azure/giantnetes/main.tf
```diff
@@ -81,14 +100,89 @@ module "vault" {
...
+  content  = "${data.ct_config.master_big.rendered}"
+  filename = "${path.cwd}/generated/master-ignition.yaml"
+}
+
+resource "azurerm_storage_account" "storage_acc" {
+  name                     = "${var.cluster_name}machine"
+  resource_group_name      = "${module.resource_group.name}"
+  location                 = "westeurope"
+  account_tier             = "Standard"
```
r7vme: also `account_kind=BlobStorage` and `access_tier=Cold` , because we need only objects and not going to host VM disks there.

platforms/azure/giantnetes/main.tf
```diff
@@ -81,14 +100,89 @@ module "vault" {
...
+  account_replication_type = "LRS"
+}
+
+resource "azurerm_storage_container" "ignition" {
+  name                  = "ignition"
+  resource_group_name   = "${module.resource_group.name}"
+  storage_account_name  = "${azurerm_storage_account.storage_acc.name}"
+  container_access_type = "container"
```
r7vme: This actually gives full anonymous access to container

modules/azure/master-as/master-as.tf
```diff
@@ -93,6 +93,11 @@ resource "azurerm_virtual_machine" "master" {
     }
   }
 
+  boot_diagnostics {
```
r7vme: Do we really need boot diagnostics? we are experts already :trollface:

calvix: Well, it gives us logs for instance if something fuck ups. If nothing breaks we don't need it. 
I am fine with removing it.

calvix: removed fro now

calvix:
moved storage to separate module

calvix:
@r7vme I took a look at container restriction and it seems not to be implemented in terraform.

this is the functionality: https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1

Here is the issue in terraform: https://github.com/terraform-providers/terraform-provider-azurerm/issues/90

And we can see that only resource quota is available in terraform: https://www.terraform.io/docs/providers/azurerm/r/storage_share.html

r7vme:
For storage access we can use following:
- `private` access mode for container
- MSI for VM with proper role ([docs](https://www.terraform.io/docs/providers/azurerm/r/virtual_machine.html#identity))

The only concern that MSI can be not available on really early stage of ignition retrieval. I think MSI smth to do with waagent, that runs later.

calvix:
Here we go, azure is green.

PR #80, opened 2018-04-26
Make logging bucket be part of a module and simplify installation steps

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/2726


examples/aws/example-build/envs.sh
```diff
@@ -27,4 +27,7 @@ export TF_VAR_root_dns_zone_id=<aws route53 zone id>
 export TF_VAR_subnet_bastion_0=<bastion subnet cidr 1>
 export TF_VAR_subnet_bastion_1=<bastion subnet cidr 2>
 
+# Logging bucket expiration days
+export TF_VAR_expiration_days=<expiration days>
+
```
pipo02mix: For the real clusters, should I set this var in conveyor? how can I get it from installations?

examples/aws/example-build/envs.sh
```diff
@@ -27,4 +27,7 @@ export TF_VAR_root_dns_zone_id=<aws route53 zone id>
 export TF_VAR_subnet_bastion_0=<bastion subnet cidr 1>
 export TF_VAR_subnet_bastion_1=<bastion subnet cidr 2>
 
+# Logging bucket expiration days
+export TF_VAR_expiration_days=<expiration days>
```
r7vme: Do we really need this here, we have a lot of variables and if we put all of them here it does not really makes sense to me. Default 365 days is fine for me.

examples/aws/example-build/envs.sh
r7vme: Just use meaningful default in variables.tf. No need to set this everywhere.

misc/e2e-aws.sh
```diff
@@ -75,6 +75,7 @@ stage-preflight() {
   [ ! -z "${E2E_AWS_SECRET_KEY+x}" ] || fail "variable E2E_AWS_SECRET_KEY is not set"
   [ ! -z "${E2E_AWS_REGION+x}" ] || fail "variable E2E_AWS_REGION is not set"
   [ ! -z "${E2E_AWS_ROUTE53_ZONE+x}" ] || fail "variable E2E_AWS_ROUTE53_ZONE is not set"
+  [ ! -z "${E2E_EXPIRATION_DAYS+x}" ] || fail "variable E2E_EXPIRATION_DAYS is not set"
```
r7vme: See comments above.

misc/e2e-aws.sh
```diff
@@ -102,6 +103,7 @@ export TF_VAR_root_dns_zone_id=${E2E_AWS_ROUTE53_ZONE}
 export TF_VAR_nodes_vault_token=
 export TF_VAR_aws_customer_gateway_id=
 export TF_VAR_worker_count=${WORKER_COUNT}
+export TF_VAR_expiration_days=${E2E_EXPIRATION_DAYS}
```
r7vme: Same here.

misc/e2e-aws.sh
```diff
@@ -223,11 +212,13 @@ stage-destroy() {
   cd ${BUILDDIR}
 
   source envs.sh
+
   terraform init ../platforms/aws/giantnetes
   terraform destroy -force ../platforms/aws/giantnetes
 
+  # Remove manually the bucket because it should stay in a prod env
```
r7vme: If it created by terraform it should be destroyed by terraform. No need to destroy it manually

pipo02mix: True now it should be done by Terraform

examples/aws/example-build/envs.sh
r7vme: Do you have this plan now? All variables available [here](https://github.com/giantswarm/giantnetes-terraform/blob/master/platforms/aws/giantnetes/variables.tf) and there are bunch of them. 
I want to avoid having "monster" size `envs.sh`. If you will need this for `viking` you can do this, but i don't think that for every new installation we should decide on logging retention period

pipo02mix: Ok so, for now, we can avoid variables, and set up as default the viking desired value. In the future if there is other client wants a different policy then we can implement. WDYT?

r7vme: let's have a call i'll describe my point :)

i want to remove it from `envs.sh` not from variables.tf

platforms/aws/giantnetes/variables.tf
```diff
@@ -24,6 +24,12 @@ variable "worker_count" {
   default     = "4"
 }
 
+variable "expiration_days" {
```
r7vme: You still need to pass this variable to s3 module

https://github.com/giantswarm/giantnetes-terraform/blob/master/platforms/aws/giantnetes/main.tf#L55-L60

pipo02mix:
Thanks for your effort reviewing!

PR #82, opened 2018-04-27
Disable conformance tests by default.

r7vme:
Makes conformance tests optional.

This is steps one to relax our experience with CI. Next step is to enable this only for master branch.

### DIFF
--- a/misc/e2e-aws.sh
+++ b/misc/e2e-aws.sh
@@ -25,6 +25,10 @@ SSH_USER="e2e"
 KUBECTL_CMD="docker run -i --net=host --rm quay.io/giantswarm/docker-kubectl:f51f93c30d27927d2b33122994c0929b3e6f2432"
 WORKER_COUNT=1
 
+# Please set any non empty value to E2E_ENABLE_CONFORMANCE in CircleCI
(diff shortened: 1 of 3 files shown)


calvix:
How do you enable/disable conformance test?
Best would be to have another `hold` button after basic e2e finish

r7vme:
We can discuss this actually. 

> How do you enable/disable conformance test?

I don't see the benefit having second hold button (no one will ever use it). We either run them together or don't bother at all. I want to have them only for master branch. To enable them i'll add variable to circleci configuration for master jobs.

misc/e2e-aws.sh
```diff
@@ -290,8 +295,8 @@ main() {
   # Wait for kubernetes nodes.
   stage-wait-kubernetes-nodes
 
-  # Finally run tests.
-  stage-e2e
+  # Finally run tests if enabled.
+  [ ! ${E2E_ENABLE_CONFORMANCE} ] || stage-e2e
```
r7vme: `!` needed to not exit with 1, of tests are disabled. Note `set -e` above it causes failure if any command returns `exit 1`.

PR #84, opened 2018-05-07
Add wait for domains

corest:
Had a race condition before while trying get ssh CA while vault address being unresolvable


corest:
Tested on azure, aws ci is currently broken

PR #85, opened 2018-05-07
fix-proxy-kubeconfig

calvix:
bugfix


calvix:
not rolling CI as this is quite straightforward and tested, also needed this ASAP :)

PR #90, opened 2018-05-15
Make instance options configurable

corest:
Towards: https://github.com/giantswarm/adidas/issues/306

I didn't tested this yet, but want to know your thoughts on this. 
For China we will need m4 instance type, so I'm making all related stuff configurable

### DIFF
--- a/platforms/aws/giantnetes/variables.tf
+++ b/platforms/aws/giantnetes/variables.tf
@@ -46,10 +46,10 @@ variable "master_instance" {
   type = "map"
 
   default = {
-    type                     = "m5.large"
-    mount_device_name_docker = "/dev/nvme1n1"
-    mount_device_name_etcd   = "/dev/nvme2n1"
-    volume_device_name_etcd  = "/dev/xvdh"
+    type         = "m5.large"
+    mount_docker = "/dev/nvme1n1"
+    mount_etcd   = "/dev/nvme2n1"
+    volume_etcd  = "/dev/xvdh"
   }
 }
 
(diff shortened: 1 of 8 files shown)


platforms/aws/giantnetes/variables.tf
```diff
@@ -42,14 +42,25 @@ variable "vault_instance_type" {
...
-  type    = "string"
-  default = "m5.large"
+variable "master_instance" {
+  type = "map"
+
+  default = {
+    type                     = "m5.large"
+    mount_device_name_docker = "/dev/nvme1n1"
+    mount_device_name_etcd   = "/dev/nvme2n1"
+    volume_device_name_etcd  = "/dev/xvdh"
```
r7vme: Naming is confusing.  Why use `mount_` and `volume_` prefixes? `device_name_xxx` pretty enough. And i think you have to remove one for `volume_device_name_etcd`.

corest: I'll adjust naming. Also `volume_device_name_etcd` - that is aws volume attachment. Naming was mostly based on the PR https://github.com/giantswarm/giantnetes-terraform/pull/76

Changes in ignition files are also there. e.g. https://github.com/giantswarm/giantnetes-terraform/pull/90/files#diff-b3bb19fde7c1bd581f41bd9d02e41b7bR2307

r7vme: I see. Then it's fine.

r7vme: I did not know get it initially

corest:
Deployed to gauss, works fine

PR #91, opened 2018-05-16
Keep data disks on worker destroy

corest:
Towards https://github.com/giantswarm/giantswarm/issues/2997

Tested on `gollum`.


corest:
Probably some CI adjustments required just in case there will be any data disks in future

calvix:
nice catch

PR #93, opened 2018-05-17
switch all images to retagged ones in our quay org

puja108:
towards giantswarm/adidas#320 goes with https://github.com/giantswarm/retagger/pull/83

I included images that were on quay but not retagged in our org


puja108:
ok, azure passed, now for aws

PR #94, opened 2018-05-21
Changes, required for China region

corest:
List of changes:

- configurable route53 records
- configurable tags for s3 buckets
- configurable IAM region (as China uses `-cn` prefix)
- configurable pod-infra-container-image


corest:
e2e passed, ptal

PR #97, opened 2018-05-25
Use cheaper resources by default

r7vme:
- use standard disks for OS (3 times cheaper)
- use standard disks for workers 
- deploy 3 workers by defualt (was 4) - tested in `gollum` looks like we have enough resources with 3 workers.
- use `v3` vm sizes (newer and cheaper)

### DIFF
--- a/platforms/azure/giantnetes/variables.tf
+++ b/platforms/azure/giantnetes/variables.tf
@@ -49,7 +49,7 @@ variable "nodes_vault_token" {
 variable "worker_count" {
   type        = "string"
   description = "Number of worker nodes to be created."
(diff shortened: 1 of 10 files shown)


platforms/azure/giantnetes/variables.tf
```diff
@@ -49,7 +49,7 @@ variable "nodes_vault_token" {
 variable "worker_count" {
   type        = "string"
   description = "Number of worker nodes to be created."
-  default     = "4"
+  default     = "3"
```
r7vme: Use 3 workers instead of 4.

PR #98, opened 2018-05-25
Use SAS for ignition object access.

r7vme:
This change makes ignition container private and creates temporary SAS credentials for 3 months. SAS is data source and it's renewed everytime terraform run. Also use md5 sum for ignition object name, this allows to prevent master recreation if no changes in ignition data. 

Fixes: https://github.com/giantswarm/giantswarm/issues/3041


r7vme:
ping @giantswarm/sre

PR #99, opened 2018-05-25
Have 3 workers in AWS

r7vme:
Scaled down one worker in `ginger` w/o  issues. So let's scale down AWS by default  too.


calvix:
I would be careful as in other installations we have bigger limits on Prometheus, ginger is probably very small in this regard.

r7vme:
Good point, i'll check in "heavy" env too. In general, even if we will not fit, as default it makes sense, we just need to adjust heavy environments in `envs.sh` (so we will have dediated node for Prom)

PR #100, opened 2018-05-28
Allow workers to access vault node exporter

tuommaki:
This is needed in order to get vault machine monitored in cloud,
prometheus needs to be allowed to scrape vault machine node-exporter.

### DIFF
--- a/modules/aws/vault/vault.tf
+++ b/modules/aws/vault/vault.tf
@@ -1,5 +1,12 @@
 data "aws_availability_zones" "available" {}
 
+# "Recreate" worker subnets in order to lookup CIDR blocks for security group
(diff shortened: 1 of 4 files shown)


tuommaki:
CI is now green. Waiting for confirmation from @r7vme now.

modules/aws/vault/vault.tf
```diff
@@ -77,6 +84,14 @@ resource "aws_security_group" "vault" {
     cidr_blocks = ["${var.vpc_cidr}"]
   }
 
+  # Allow node-exporter from worker nodes.
```
r7vme: Isn't this duplicate of `azurerm_network_security_rule` below?

PR #102, opened 2018-06-07
Allow guest nodes access to vault

corest:
Towards https://github.com/giantswarm/adidas/issues/344


fgimenez:
To give a bit more context, we are currently using KMS for encrypting/decrypting the cloudconfigs stored in s3 that are required to bootstrap guest instance nodes. In China regions KMS is not available, one of the options to replace it is using Vault. In order to use it, guest instances need to access Vault to be able to decrypt the cloudconfigs, just as they access KMS where it is available. In this initial approach, we decided to reuse the same Vault instance already created in the host cluster so that we don't need to provision and maintain additional "pets".

PR #103, opened 2018-06-13
increase-kubelet-start-timeout

calvix:
in Atlantis and axolotl the pull of images is quite slow and killing kubelet in the process is slowing the pull even more

We did similar fix also for guest cluster


calvix:
azure is weird today

calvix:
(CI was green finally)

PR #104, opened 2018-06-19
Split access between different users

corest:
Not sure about the approach, because in hive we have templating so I implemented scopes there. But here we just have pure ignition so I don't know another way to restrict access. Thoughts are welcome


corest:
@r7vme the reason for this splitting is that we still need provide access to bastions at least so users can get to the vault somehow for requesting ssh cert. Also, we don't need it per clusters. It is the same for all clusters. 
The only possible solution to make it in one file - template it somehow with internal tooling based on scopes as we have in hive. Going further, we can something like `opsct create users -r <repo>` with implementation per repository for proper templating

r7vme:
Moving ssh keys this from this repo still makes sense.

You don't need templating imo. Just having `bastion-users.yaml` `users.yaml` should be enough

PR #106, opened 2018-06-21
add add-eviction-policies

calvix:
to align with guest cluster https://github.com/giantswarm/k8scloudconfig/pull/390


calvix:
azure failed because of the missing backups env, aws is green and the change is same in both envs

merging

PR #109, opened 2018-06-26
Wipe docker disk

r7vme:
Fixes https://github.com/giantswarm/giantswarm/issues/3475


calvix:
We already do that on KVM

PR #111, opened 2018-06-27
Add VPN support

r7vme:
- Create VPN gateway and other required resources
- Manage shared key externally (users creates VPN connection resource with shared key on it's own)

### DIFF
--- a/docs/installation-guide-azure.md
+++ b/docs/installation-guide-azure.md
@@ -81,6 +81,14 @@ export ARM_ENVIRONMENT="german"
 export TF_VAR_azure_cloud=AZUREGERMANCLOUD
 ```
 
+Optionally for VPN support add following variables. `bastion_cidr` should be unique and a part of `vnet_cidr` (10.0.0.0/16 by default). Recommended to use /28 subnets from range 10.0.4.0/22 (e.g. 10.0.4.0/28, 10.0.4.16/28, etc.).
+```
+export TF_VAR_vpn_enabled=1
+export TF_VAR_vpn_right_gateway_address=<ip address of IPSec server>
+export TF_VAR_vpn_right_subnet_cidr=<subnet that will be shared by IPSec server>
+export TF_VAR_bastion_cidr=<bastion subnet>
+```
+
 ```
 source envs.sh
 ```
@@ -111,6 +119,24 @@ terraform apply ../platforms/azure/giantnetes
 ```
 
 It should create all cluster resources. Please note master and worker vms are created, but will fail. This is expected behaviour.
+
+#### (Optional) Connect to VPN
+
+If VPN enabled, two additional manual steps are required:
+1. Create Azure VPN connection with shared key.
+2. Create new IPSec connection in onpremise VPN server.
+
+For step one execute following command.
+```
+az network vpn-connection create \
+  -g ${NAME} \
+  --name ${NAME}-vpn-connection \
+  --vnet-gateway1 ${NAME}-vpn-gateway \
+  --local-gateway2 ${NAME}-vpn-right-gateway \
+  --shared-key <put_your_shared_key_here>
+```
+
+Step two is individual and depends on your setup.
 
 #### Provision Vault with Ansible
 
--- a/modules/azure/dns/dns.tf
+++ b/modules/azure/dns/dns.tf
@@ -40,21 +40,7 @@ resource "azurerm_dns_ns_record" "dns_zone_propagation" {
   resource_group_name = "${var.root_dns_zone_rg}"
   ttl                 = 300
 
-  record {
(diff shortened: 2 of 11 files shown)


modules/azure/dns/dns.tf
```diff
@@ -40,21 +40,7 @@ resource "azurerm_dns_ns_record" "dns_zone_propagation" {
...
-  }
-
-  record {
-    nsdname = "${azurerm_dns_zone.dns_zone.name_servers[2]}"
-  }
-
-  record {
-    nsdname = "${azurerm_dns_zone.dns_zone.name_servers[3]}"
-  }
+  records = ["${azurerm_dns_zone.dns_zone.name_servers}"]
```
r7vme: Because `record` is deprecated.

modules/azure/vnet/nic-bastion.tf
```diff
@@ -9,12 +9,12 @@ resource "azurerm_network_interface" "bastion" {
     private_ip_address_allocation = "dynamic"
     name                          = "${var.cluster_name}-bastionIPConfiguration"
     subnet_id                     = "${element(azurerm_subnet.bastion_subnet.*.id, count.index)}"
-    public_ip_address_id          = "${element(azurerm_public_ip.bastion_public_ip.*.id, count.index)}"
+    public_ip_address_id          = "${var.vpn_enabled ? "" : element(concat(azurerm_public_ip.bastion_public_ip.*.id, list("")), count.index)}"
```
r7vme: this `concat` `list` is just workaround, becase element does not support empty list.

docs/installation-guide-azure.md
```diff
@@ -81,6 +81,14 @@ export ARM_ENVIRONMENT="german"
 export TF_VAR_azure_cloud=AZUREGERMANCLOUD
 ```
 
+Optionally for VPN support add following variables. `bastion_cidr` should be unique and a part of `vnet_cidr` (10.0.0.0/16 by default). Recommended to use /28 subnets from range 10.0.4.0/22 (e.g. 10.0.4.0/28, 10.0.4.16/28, etc.).
```
corest: Shouldn't we reference this page here ? https://github.com/giantswarm/giantswarm/wiki/Giant-Swarm-VPN 
Bastion cidr should be unique across that table

modules/azure/vpn/vpn.tf
```diff
@@ -0,0 +1,40 @@
...
+
+  type     = "Vpn"
+  vpn_type = "RouteBased"
+
+  active_active = false
+  enable_bgp    = false
+
+  # SKU details can be found here
+  # https://docs.microsoft.com/en-us/azure/vpn-gateway/vpn-gateway-about-vpngateways
+  sku = "VpnGw1"
```
r7vme: `Basic` only supports 10 tunnels, so use `VpnGw1` which supports up to 30 tunnels, because we will use it for BYOC guest clusters.

docs/installation-guide-azure.md
r7vme: I'm trying to avoid referencing any internal repos / wiki , because it's open source repo.

corest: Ah, ok

platforms/azure/giantnetes/variables.tf
```diff
@@ -159,6 +159,15 @@ variable "vnet_cidr" {
   default     = "10.0.0.0/16"
 }
 
+# NOTE:
+# - bastion_cidr should be a part of vnet_cidr.
+# - bastion_cidr should be unique across all installations if VPN enabled.
+# - recommended to use /28 subnets from range 10.0.4.0/22 (for default 10.0.0.0/16 vnet_cidr).
```
TheoBrigitte: Just for my information, did you understood where this recommendation of /28 came from ?

r7vme: It's my recommendation, so we can have this as baseline. No strict policy tho. `/28` is used in AWS.

Any requirements from BYOC side?

r7vme: For reference https://github.com/giantswarm/giantswarm/wiki/Giant-Swarm-VPN

TheoBrigitte: Ok, for info I made it /24 on guest cluster because why not? and I could not find a reason not to.

r7vme: `/24` is also fine. In guest clusters we already by design have non-intersecting networks, so we just fine with any range.

In VPN server <=> host clusters we should be careful, because we use almost always the same range for host cluster `10.0.0.0/16` That's why i'm recommending to take `10.0.4.0/22` as range and pick up `/28` networks from it. This give us 64 potential host clusters. And we always can pick up another range if this one will be occupied.

PR #117, opened 2018-07-06
Setup two vpn connection

corest:
Towards https://github.com/giantswarm/giantswarm/issues/3534


corest:
Ok, this approach will not work as we can have only one Virtual gateway, which can be bounded to the VPC

corest:
Tested this on gauss, works fine

platforms/azure/giantnetes/main.tf
```diff
@@ -244,14 +244,26 @@ module "worker" {
...
   resource_group_name       = "${module.resource_group.name}"
   subnet_id                 = "${module.vnet.vpn_subnet_id}"
   vpn_enabled               = "${var.vpn_enabled}"
-  vpn_right_gateway_address = "${var.vpn_right_gateway_address}"
-  vpn_right_subnet_cidr     = "${var.vpn_right_subnet_cidr}"
+  vpn_right_gateway_address = "${var.vpn_right_gateway_address_0}"
+  vpn_right_subnet_cidr     = "${var.vpn_right_subnet_cidr_0}"
+}
+
+module "vpn" "vpn_connection_1" {
```
r7vme: We either:
1. call module twice (like here) and module uses some suffixes for resources it creates. But we need to remove hard coding with `_0` `_1` from inside module (module does not know that there are two VPNs at the same time).
2. or just call module once and always have two gateways (module by default creates two vpn gateways)

I like the first one as this is actual purpose of modules.

Currently you have this knowledge (about two vpns) both inside module and when calling module.

corest:
I’ve tried first approach, but it looks even uglier than second. Also in treating this in the same way as subsets for ELB, that is just mandatory HA

corest:
ping @giantswarm/sre

platforms/aws/giantnetes/main.tf
```diff
@@ -243,10 +242,12 @@ module "vpn" {
   source = "../../../modules/aws/vpn"
 
   # If aws_customer_gateway_id is not set, no vpn resources will be created.
```
r7vme: nitpick edit in comment too `aws_customer_gateway_id_0`

platforms/aws/giantnetes/variables.tf
```diff
@@ -210,16 +210,26 @@ variable "subnet_worker_1" {
...
 
-variable "aws_customer_gateway_id" {
+variable "aws_customer_gateway_id_0" {
   description = "AWS customer gateway ID. Should be created manually. Leave blank to disable VPN setup and enable public access for bastions."
   default     = ""
 }
 
-variable "external_ipsec_subnet" {
+variable "aws_customer_gateway_id_1" {
+  description = "AWS customer gateway ID. Should be created manually. Leave blank to disable VPN setup and enable public access for bastions."
```
r7vme: nitpick, remove `Leave blank to disable VPN setup and enable public access for bastions.` as this is only true for `_0`

PR #119, opened 2018-07-11
Mitigate issue with Azure disk flip

r7vme:
Looks like in Azure disk also flipping. This is just a mitigation.

Mitigate: https://github.com/giantswarm/giantswarm/issues/3700


r7vme:
>  But why does this random pick happen?

No idea atm, there is a postmortem for that.

PR #120, opened 2018-07-18
Default priorities for k8s 1.11

corest:
Towards:
https://github.com/giantswarm/giantswarm/issues/2445
https://github.com/giantswarm/giantswarm/issues/3618

So, in k8s 1.11 priorities are enabled by default. 
There are two default priority classes `system-node-critical`, `system-cluster-critical`, which are used for our critical components. Calico and kube-proxy do not have this as they are not controlled by default scheduler.

### DIFF
--- a/ignition/aws/worker.yaml.tmpl
+++ b/ignition/aws/worker.yaml.tmpl
@@ -46,6 +46,29 @@ storage:
               user: kubelet
             name: service-account-context
           current-context: service-account-context
+    - path: /etc/kubernetes/config/kubelet-config.yaml.tmpl
+      filesystem: root
+      mode: 0644
+      contents:
+        inline: |
+          kind: KubeletConfiguration
+          apiVersion: kubelet.config.k8s.io/v1beta1
+          address: $${DEFAULT_IPV4}
+          port: 10250
+          healthzBindAddress: $${DEFAULT_IPV4}
+          healthzPort: 10248
+          clusterDNS: ${K8S_DNS_IP}
+          clusterDomain: cluster.local
+          featureGates:
+            ExpandPersistentVolumes: true
+          podManifestPath: /etc/kubernetes/manifests
+          evictionSoft:
+            memory.available:  "500Mi"
+          evictionHard:
+            memory.available:  "200Mi"
+          evictionSoftGracePeriod:
+            memory.available:  "5s"
+          evictionMaxPodGracePeriod: 60
     - path: /etc/kubernetes/config/proxy-config.yaml
       filesystem: root
       mode: 0644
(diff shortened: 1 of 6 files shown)


corest:
There is a section in default scheduler-config, which I removed for cloud-clusters as we don't have multi-masters. This will be used in on-prem only

```
leaderElection:
  leaderElect: true
  leaseDuration: 15s
  lockObjectName: kube-scheduler
  lockObjectNamespace: kube-system
  renewDeadline: 10s
  resourceLock: endpoints
  retryPeriod: 2s
```

ignition/aws/worker.yaml.tmpl
```diff
@@ -46,6 +46,48 @@ storage:
...
+          bindAddress: 0.0.0.0
+          clientConnection:
+            acceptContentTypes: ""
+            burst: 10
+            contentType: application/vnd.kubernetes.protobuf
+            kubeconfig: /etc/kubernetes/config/proxy-kubeconfig.yaml
+            qps: 5
+          clusterCIDR: ""
+          configSyncPeriod: 15m0s
+          conntrack:
```
r7vme: Do we really need all these settings? Can they be defaults?

ignition/azure/master.yaml.tmpl
```diff
@@ -1460,6 +1463,71 @@ storage:
...
+    - path: /etc/kubernetes/config/scheduler-config.yaml
+      filesystem: root
+      mode: 0644
+      contents:
+        inline: |
+          kind: KubeSchedulerConfiguration
+          algorithmSource:
+            provider: DefaultProvider
+          apiVersion: componentconfig/v1alpha1
+          clientConnection:
```
r7vme: Same here

ignition/azure/master.yaml.tmpl
```diff
@@ -1460,6 +1463,71 @@ storage:
               user: proxy
             name: service-account-context
           current-context: service-account-context
+    - path: /etc/kubernetes/config/scheduler-config.yaml
```
r7vme: Why we don't have kubeproxy config for Azure? Files for AWS and Azure should be almost the same imo, except calico.

ignition/azure/worker.yaml.tmpl
```diff
@@ -399,7 +399,7 @@ systemd:
       RestartSec=0
       TimeoutStopSec=10
       EnvironmentFile=/etc/network-environment
-      Environment="IMAGE=quay.io/giantswarm/hyperkube:v1.10.2"
+      Environment="IMAGE=quay.io/giantswarm/hyperkube:v1.11.1"
```
r7vme: Same here changes from aws worker.yaml should be here too, right?

ignition/aws/worker.yaml.tmpl
corest: That is how default config file looks

ignition/azure/master.yaml.tmpl
corest: That is how default config file looks

corest:
Deployed this to gauss. Seems to be fine

ignition/aws/worker.yaml.tmpl
r7vme: But why do we need everything specified here? can we only have settings that we override in config?

corest:
That step can't be moved into docker. It is part of ansible bootstrap role and related to the outdated libncurse version. Testing fix now

corest:
@giantswarm/sre hey folks, can you check last commit? On aws I had an issue, that `*-certs` services failed because of failed lookup on vault address. That is kind of strange, taking into account that there is pre-check for DNS. Turned out, that for container it takes another amount of time to resolve address properly. That's why I switch curl from host check into the container check. That finally solved broken CI.
wdyt?

r7vme:
really strange, both inside and outside container same DNS server (AWS internal) will be used...

corest:
True, but that is what happening. If no veto, I’ll merge this tomorrow

PR #121, opened 2018-07-20
add conveyor ssh user

calvix:
add conveyor user to bastions
 towards https://github.com/giantswarm/giantswarm/issues/3806


calvix:
@r7vme AFAIk we need users on bastion in order to issue cert for SSO, same for vault

PR #122, opened 2018-07-26
unify-lb-health-checks

calvix:
towards https://github.com/giantswarm/giantswarm/issues/3208


calvix:
tested on ghost and all seem to be fine

PR #123, opened 2018-07-26
Identify Azure disk by LUN

r7vme:
part of: https://github.com/giantswarm/giantswarm/issues/3582

### DIFF
--- a/ignition/azure/master.yaml.tmpl
+++ b/ignition/azure/master.yaml.tmpl
@@ -1859,11 +1859,48 @@ systemd:
   - name: systemd-networkd-wait-online.service
     enabled: false
     mask: true
+  - name: format-docker-disk.service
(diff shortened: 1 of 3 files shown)


r7vme:
This does not work. See upstream issue https://github.com/coreos/bugs/issues/2481

Problem that ignition does not have Azure specific udev rules in early user space.

We will need to move back to traditional systemd units to format and mount fs.

r7vme:
Tested in `ghost` works fine. Will merge after CI.

r7vme:
Seems are good news, CoreOS communnity confirmed a bug https://github.com/coreos/bugs/issues/2481

Closing this PR (as we have mitigation for now) and just wait until bug above will be fixed, so we can continue using "filesystems". pew pew

PR #125, opened 2018-07-26
Upgrade nginx version to 0.16.2

pipo02mix:
Towards https://github.com/giantswarm/giantswarm/issues/3717

I have tested this version in AWS and Azure host clusters. Even [release page describes](https://github.com/kubernetes/ingress-nginx/releases/tag/nginx-0.16.2) a breaking change it does not affect since we don't apply any psp to IC in host clusters. Adding the security context at container level in our case make the deployment fails (because cannot validate against any PSP)


calvix:
would like to see passing e2e before merge

PR #127, opened 2018-08-02
Associate bastions into private routing table

corest:
As part of enabling logs on bastions, internet access should be turned on there. Also, as bastions don't have now public addresses, they can be associated with private routing table.

  - switch nat gateways to elb subnets (which are the only in public now)
  - move IPSec routes from public routing table into private, which is
  now used by bastions


corest:
fixed, ptal

corest:
Added separate issue as this one blocks me

PR #128, opened 2018-08-03
Forward bastion logs to cloudwatch

corest:
Towards https://github.com/giantswarm/giantswarm/issues/3297

Default log priority is "none" which means service will be disabled. 
To enable it, override terraform variable `TF_VAR_bastion_log_priority`, eg `export TF_VAR_bastion_log_priority=err`. Supported values are the same as priorities for `journald`


corest:
The bastion ignition split is temporary. As next task I'll write templating for ignition

Issue #130, opened 2018-08-07
Generate terraform ignition from templates


corest:

corest:
Done

PR #131, opened 2018-08-07
Ignition rendering poc

corest:
Towards: https://github.com/giantswarm/giantnetes-terraform/issues/130

We have now separate ignition files for aws/azure bastion/vault/master/worker nodes. So when we are changing something, it usually goes into both files. This creates some overhead in management as a lot of data is just duplicated. 

This is an attempt to solve this. The idea is to generate actual ignition files (which are rendered with vaules by terraform) via go templates. For this approach small utility is used (https://github.com/giantswarm/ignition-renderer/blob/renderer/main.go). Whenever template is changed - it is re-rendered for the proper platform (aws/azure) and then passed to the terraform. 

This PR is poc for managing bastion hosts. And it removes ~150 lines of duplicated ignition with templating.

This also has its disadvantages. With the single template, changes for one platform will cause terraform reapply for another platform. 

So, the question is do we really want this templating?


calvix:
I like this, this makes sense to me, but maybe we can have the rendering tool as a terraform plugin which would simplify installing ( as terraform will do it for us). We do it similarly with ct plugin https://github.com/coreos/terraform-provider-ct

from my side LGTM if it works properly.

corest:
@calvix I've looked into terraform plugins. It is much more complex and would be worth to do it in case of complex configs. In our case, we are using only one parameter(aws/azure) and that seems to be too simple for plugin. 

Regarding simplifying installing - with binary in the repository - there is no need to install anything additional.

corest:
@r7vme 
Tried the plugin you've mentioned and I think we are on a good way with removing all the hacks

corest:
Deployed to gauss and now deploying to azure. Seems to be good. There is only step for generating ignition from template and then for converting to cloudconfig.

ptal pls

templates/worker.yaml.tmpl
```diff
@@ -487,9 +493,14 @@ systemd:
       --containerized \
       --enable-server \
       --logtostderr=true \
+      {{if eq .Provider "azure" -}}
+      --cloud-provider=azure \
+      --cloud-config=/etc/kubernetes/config/azure.yaml \
+      {{ else -}}
       --cloud-provider=aws \
+      --pod-infra-container-image={{ .PodInfraImage }} \
```
calvix: isn't this more like general things rather than aws specific?

.circleci/config.yml
```diff
@@ -24,7 +24,7 @@ jobs:
 
   e2eTestAWS: &e2eTestAWS
     docker:
-      - image: quay.io/giantswarm/docker-terraform-and-stuff:latest
+      - image: quay.io/giantswarm/docker-terraform-and-stuff:gotemplate
```
calvix: You probably want to switch back to latest before the merge.

.circleci/config.yml
```diff
@@ -36,7 +36,7 @@ jobs:
 
   e2eTestAzure: &e2eTestAzure
     docker:
-      - image: quay.io/giantswarm/docker-terraform-and-stuff:latest
+      - image: quay.io/giantswarm/docker-terraform-and-stuff:gotemplate
     steps:
     - checkout
     - run: ./misc/e2e-azure.sh
```
calvix: You probably want to switch back to latest before the merge.

.circleci/config.yml
```diff
@@ -36,7 +36,7 @@ jobs:
 
   e2eTestAzure: &e2eTestAzure
     docker:
-      - image: quay.io/giantswarm/docker-terraform-and-stuff:latest
+      - image: quay.io/giantswarm/docker-terraform-and-stuff:gotemplate
```
calvix: You probably want to switch back to latest before the merge.

calvix:
looks much better than before, LGTM IFWY

.circleci/config.yml
corest: sure

templates/worker.yaml.tmpl
corest: Yes, it is general. But used only on AWS as we are overriding default for China

PR #133, opened 2018-08-16
Change terminology from host cluster to control plane

oponder:
Towards: https://github.com/giantswarm/giantswarm/issues/3991


r7vme:
Let's merge it @oponder :)

oponder:
@r7vme: I want to! But I dont know what kind of craziness will happen if I do merge this. Also theres the e2e test on hold? Should I let it run?

r7vme:
All good, nothing will happen. This is docs only.

PR #134, opened 2018-08-21
Add kernel modules required for IPVS

corest:
Towards https://github.com/giantswarm/giantswarm/issues/3617

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -2294,6 +2294,16 @@ storage:
           SUBSYSTEM=="block", KERNEL=="sdc", TAG+="systemd"
           SUBSYSTEM=="block", KERNEL=="sda6", TAG+="systemd"
     {{- end }}
+    - path : /etc/modules-load.d/ipvs.conf
+      filesystem: root
+      mode: 0644
+      contents:
+        inline: |
+          ip_vs
+          ip_vs_rr
+          ip_vs_wrr
+          ip_vs_sh
+          nf_conntrack_ipv4
 systemd:
   units:
   - name: etcd2.service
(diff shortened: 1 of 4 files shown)


templates/master.yaml.tmpl
```diff
@@ -2294,6 +2294,16 @@ storage:
...
+    - path : /etc/modules-load.d/ipvs.conf
+      filesystem: root
+      mode: 0644
+      contents:
+        inline: |
+          ip_vs
+          ip_vs_rr
+          ip_vs_wrr
+          ip_vs_sh
+          nf_conntrack_ipv4
```
r7vme: Do we really need `nf_conntrack_ipv4`? it's loaded by default afaik

corest: I just took a list of modules from official k8s blog post. Maybe it is not required there, but it is idempotent from the CoreOS version

r7vme:
One more question. Do we have idea on runtime migration?

1. what if kube-proxy with ipvs will run on worker w/o modules (because we roll master node first) ?
2. what happen when both ipvs and iptables kube-proxy are running in the cluster? should be fine i think as we have all rules on every node.

templates/master.yaml.tmpl
r7vme: oki :)

r7vme:
> kube-proxy daemonset will be updated with flag, which overrides iptables mode

Is it some specific compatibility mode?

corest:
No, kubeproxy requires configfile, which is generated from ignition. And if we put ipvs into config file - ds will fail after master upgrade as all workers will not have yet new configfile.

r7vme:
ack. now it's clear.

Would be great to document this in the ticket, where we track IPVS migration. Because seems in "tenant clusters" it will take 3 releases before fully moved to final state.

release 1 - with kernel modules
release 2 - with kube-proxy flag - IPVS enabled at this point
release 3 - with kube-proxy with config file

correct?

PR #142, opened 2018-09-02
Use longer image pull progress deadline for AWS China regions

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/3936


rossf7:
CI is failing so I must have screwed something up. I'll debug this next week but will likely need some help as my knowledge here is shallow.

rossf7:
Thanks for the feedback. The problem is just with the worker templating. If I remove it CI is fine but with it e2e fails because the worker node isn't ready.

I've updated so the conditional logic is the same for master and workers. Re-running e2e now.

PR #143, opened 2018-09-10
Remove sso from bastions

corest:
As we anyway keep user accounts on bastions, we can remove sso ssh from those nodes

### DIFF
--- a/templates/bastion.yaml.tmpl
+++ b/templates/bastion.yaml.tmpl
@@ -16,6 +16,16 @@ systemd:
     - name: fleet.service
       enabled: false
       mask: true
+    - name: os-hardening.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Apply os hardening
+        [Service]
+        Type=oneshot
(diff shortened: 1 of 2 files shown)


templates/bastion.yaml.tmpl
```diff
@@ -16,64 +16,6 @@ systemd:
...
-        [Unit]
-        Description=Wait for etcd and k8s API domains to be available
-
-        [Service]
-        Type=oneshot
-        ExecStart=/opt/wait-for-domains
-
-        [Install]
-        WantedBy=multi-user.target
-    - name: os-hardening.service
```
r7vme: os-hardening should be present imo

templates/bastion.yaml.tmpl
```diff
@@ -103,12 +45,6 @@ systemd:
     {{end }}
 storage:
   files:
-    - path: /etc/tokens/node
-      filesystem: root
-      mode: 0400
-      contents:
-        inline: |
-          VAULT_TOKEN={{ .G8SVaultToken }}
     - path: /etc/ssh/sshd_config
```
r7vme: I this we also can get rid of this one?

corest:
fixed, ptal

PR #144, opened 2018-09-11
Create proper calico certificates

r7vme:
Part of: https://github.com/giantswarm/giantswarm/issues/4173

This is only step 1 into migrating to proper calico certificates. This change makes sure we have `/etc/kubernetes/ssl/calico/` with `etcd-ca` `etcd-cert` `etcd-key`. At the moment we use ones from `/etc/kubernetes/ssl/etcd` and they passed (to CNI) via [environment variables](https://github.com/giantswarm/giantnetes-terraform/blob/master/templates/master.yaml.tmpl#L2852-L2854) in k8s-kubelet.service, which is not right.

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -590,6 +590,9 @@ storage:
 
             # If you're using TLS enabled etcd uncomment the following.
             # You must also populate the Secret below with these files.
+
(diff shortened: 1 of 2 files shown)


calvix:
this will probably need lots of testing :p ping when all is green.

r7vme:
Please PTAL

PR #145, opened 2018-09-12
Improve AWS ASG scaling settings

r7vme:
Part of: https://github.com/giantswarm/giantswarm/issues/4168

Align with approach we use in aws-operator. So while upgrade we scale up on 1 VM and then start rolling others.

Tested in `ginger`

### DIFF
--- a/modules/aws/worker-asg/worker-asg.tf
+++ b/modules/aws/worker-asg/worker-asg.tf
@@ -24,8 +24,9 @@ resource "aws_cloudformation_stack" "worker_asg" {
         "LoadBalancerNames": [
           "${var.cluster_name}-worker"
         ],
-        "MaxSize": "${var.worker_count}",
-        "MinSize": "1",
+        "MaxSize": "${var.worker_count + 1}",
+        "DesiredCapacity": "${var.worker_count}",
+        "MinSize": "${var.worker_count}",
         "Tags": [
           {
             "Key": "Name",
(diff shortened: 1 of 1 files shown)


modules/aws/worker-asg/worker-asg.tf
```diff
@@ -24,8 +24,9 @@ resource "aws_cloudformation_stack" "worker_asg" {
         "LoadBalancerNames": [
           "${var.cluster_name}-worker"
         ],
-        "MaxSize": "${var.worker_count}",
-        "MinSize": "1",
+        "MaxSize": "${var.worker_count + 1}",
+        "DesiredCapacity": "${var.worker_count}",
+        "MinSize": "${var.worker_count}",
```
kopiczko: Does that mean we'll run one more instance than specified? Isn't it more explicit to just add one node to all AWS control planes?

tuommaki: I understand this that `min == number of workers normally`, `desired == min` and `max == desired + 1` which allows to scale by one during updates?

r7vme: >  Isn't it more explicit to just add one node to all AWS control planes?

It's more expensive to always run 4 VMs.

> I understand this that min == number of workers normally, desired == min and max == desired + 1 which allows to scale by one during updates?

Yes.

kopiczko: Ah, yes. Me dumb.

PR #148, opened 2018-09-24
update-api-args

calvix:
towards https://github.com/giantswarm/giantswarm/issues/4211

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -2193,8 +2193,6 @@ storage:
                 - /hyperkube
                 - apiserver
                 - --allow-privileged=true
-                - --insecure-bind-address=127.0.0.1
-                - --insecure-port=8080
                 - --kubelet-https=true
                 - --secure-port=443
                 {{ if eq .Provider "aws" -}}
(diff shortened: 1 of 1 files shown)


templates/master.yaml.tmpl
```diff
@@ -2206,7 +2204,7 @@ storage:
                 - --bind-address=$(HOST_IP)
                 - --etcd-prefix=giantswarm.io
                 - --authorization-mode=RBAC
-                - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota,PodSecurityPolicy,PersistentVolumeClaimResize,DefaultStorageClass,Priority
+                - --enable-admission-plugins=NamespaceLifecycle,LimitRanger,ServiceAccount,ResourceQuota,PodSecurityPolicy,PersistentVolumeClaimResize,DefaultStorageClass,Priority,DefaultTolerationSeconds,MutatingAdmissionWebhook,ValidatingAdmissionWebhook
```
r7vme: Do you know why we enable those 3 admission plugins?

calvix: https://github.com/giantswarm/giantswarm/issues/4211#issuecomment-422751994
@pipo02mix wanted them so i assume he has reasons

r7vme: ack

pipo02mix: I have arrived late. Yes we want to offer the possibility to configure admission webhooks (validation/mutation) to customers. For version 1.10 and newer they are recommended to be configured as default https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers#is-there-a-recommended-set-of-admission-controllers-to-use

PR #149, opened 2018-09-26
add info about isntalling plugins

calvix:
add info how install terraform plugins


corest:
There is this kind of info already here : https://github.com/giantswarm/giantnetes-terraform/blob/master/docs/installation-guide-aws.md#appendix-a So If we're adding it into main README.md, probably we can remove it from aws-specific readme

calvix:
Right, make sense

PR #152, opened 2018-09-27
Adjust terraform for china

calvix:
* make forwarding log to cloudwatch optional - china doesn't support cloudwatch vpc endpoint
* make docker registry configurable - quay.io in China is nightmare using china registry speed up everything a lot
* disable route53.public resource when route53 is disabled

before merging we need to  remove registry value from pod infra container in installations where its set

### DIFF
--- a/misc/e2e-azure.sh
+++ b/misc/e2e-azure.sh
@@ -207,6 +207,8 @@ bare_metal: False
 EOF
 
     # Bootstrap insecure Vault.
+    echo "waiting for vault node 6min"
+    sleep 6m
     export ANSIBLE_HOST_KEY_CHECKING=False
     ansible-playbook -i hosts_inventory/${CLUSTER} -e dc=${CLUSTER} bootstrap1.yml
 
@@ -254,7 +256,7 @@ stage-debug() {
 
 stage-destroy() {
   stage-debug || true
-  sleep 8m
+  
   cd ${BUILDDIR}
   source envs.sh
   terraform init ../platforms/azure/giantnetes
--- a/misc/e2e-azure.sh
+++ b/misc/e2e-azure.sh
@@ -241,7 +241,7 @@ stage-debug() {
 
 stage-destroy() {
   stage-debug || true
-
+  sleep 15m
   cd ${BUILDDIR}
   source envs.sh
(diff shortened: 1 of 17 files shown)


modules/aws/bastion/bastion-logs.tf
```diff
@@ -1,12 +1,14 @@
 resource "aws_cloudwatch_log_group" "bastion_log_group" {
-  name = "${var.cluster_name}_bastion"
+  count = "${ var.forward_logs_enabled ? 1 : 0 }"
```
r7vme: nitpick `${var.forward_logs_enabled ? 1 : 0 }`

because at some places you use with whitespace at some w/o

modules/aws/bastion/bastion-logs.tf
```diff
@@ -1,12 +1,14 @@
...
+  name  = "${var.cluster_name}_bastion"
 
   tags = {
     "giantswarm.io/installation" = "${var.cluster_name}"
   }
 }
 
 resource "aws_cloudwatch_log_stream" "bastion_logs" {
+  count          = "${ var.forward_logs_enabled ? 1 : 0 }"
```
r7vme: same

modules/aws/bastion/bastion.tf
```diff
@@ -130,6 +130,7 @@ data "ignition_config" "s3" {
 }
 
 resource "aws_vpc_endpoint" "cloudwatch" {
+  count              = "${ var.forward_logs_enabled ? 1 : 0 }"
```
r7vme: same

calvix:
aws green, but azure is doing some crazy thing

misc/e2e-azure.sh
```diff
@@ -241,7 +254,7 @@ stage-debug() {
 
 stage-destroy() {
   stage-debug || true
-
+  sleep 8m
```
kopiczko: sleep 8m?

calvix: trying to debug why azure is failing

calvix:
@giantswarm/sre so after adding `sleep 6m` before running vault on azure e2e all is green. WDYT should i leave the sleep there or remove it?

PR #153, opened 2018-10-04
Enable ipvs proxymode in kube-proxy

corest:
Towards https://github.com/giantswarm/giantswarm/issues/3617

Second stage


corest:
I've enabled it on gauss. Will keep this opened for the time of testing

PR #156, opened 2018-10-08
Add cni plugin limits to get Guaranteed QoS policy

r7vme:
Add limits to `cni` container too, this will make QoS `Guaranteed` for calico.


r7vme:
Closing as install-cni requires more than 50MB RAM and it's only needs it for few seconds and then idling.

PR #158, opened 2018-10-12
add json date of last update to route53 zone for CI cleaner

calvix:
add some data to aws route53 zone comment to found out when it was created for CI cleaner

for https://github.com/giantswarm/ci-cleaner/pull/21

### DIFF
--- a/modules/aws/dns/dns.tf
+++ b/modules/aws/dns/dns.tf
@@ -23,6 +23,10 @@ resource "aws_route53_zone" "public" {
     Name                         = "${var.zone_name}"
     "giantswarm.io/installation" = "${var.cluster_name}"
   }
+
+  lifecycle {
+    ignore_changes = ["comment"]
+  }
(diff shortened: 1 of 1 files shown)


calvix:
it looks like this
```
{"last_updated":"2018-10-12T08:31:52Z","managed_by":"terraform"}
```

modules/aws/dns/dns.tf
```diff
@@ -15,13 +15,18 @@ variable "zone_name" {
...
+  name    = "${var.zone_name}"
 
   tags {
     Name                         = "${var.zone_name}"
     "giantswarm.io/installation" = "${var.cluster_name}"
   }
+
+  lifecycle {
+    ignore_changes = ["comment"]
+  }
```
calvix: we ignore comment values for computing changes, otherwise each run of TF would update the dns zone which does not make sense

PR #160, opened 2018-10-16
Warning for control-plane nodes

corest:
1. Ask before reboot/shutdown
2. Modify PS1 with additional info 

towards: https://github.com/giantswarm/giantswarm/issues/4407

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -2400,15 +2400,15 @@ storage:
       mode: 0444
       contents:
         inline: |
-          alias shutdown="/usr/local/bin/confirm /usr/sbin/shutdown"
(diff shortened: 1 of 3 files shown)


corest:
If talk only about the issue - yes. But to keep this aligned - I would have the same here. Also, restarting accidentally any node in control-plane not the thing we want to do often?

r7vme:
Then it raises the question, why we don't have this tenant clusters, which are more critical as they run customer workloads?

corest:
The issue is scoped for the control-plane. If we have will to restart tenant node - that is usually conscious enough.  And restarting of the control-plane nodes is something that can happen while accessing tenant node

PR #162, opened 2018-10-18
Enable ipvs in kube-proxy

corest:
I didn't find issues I had before. Rolled this out to gauss/gollum - also didn't get any issues


corest:
I've discovered  that IPVS issue in control-plane is not raising after ipvs is there. Is it is just one kube-proxy restart after move. Second upgrade works fine.

JosephSalisbury:
I wanna see network latency before and after :D

corest:
Seems to be good after fixing calico

PR #164, opened 2018-10-19
Fix calico-node vs kube-proxy race for iptables lock

corest:
Towards https://github.com/giantswarm/giantswarm/issues/4395

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -300,6 +300,9 @@ storage:
                       - mountPath: /lib/modules
                         name: lib-modules
                         readOnly: true
+                      - mountPath: /run/xtables.lock
+                        name: xtables-lock
+                        readOnly: false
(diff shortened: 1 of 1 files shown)


templates/master.yaml.tmpl
```diff
@@ -300,6 +300,9 @@ storage:
                       - mountPath: /lib/modules
                         name: lib-modules
                         readOnly: true
+                      - mountPath: /run/xtables.lock
+                        name: xtables-lock
+                        readOnly: false
```
calvix: isnt this file? if you try mount file that is not there then docker will create folder instead of it
AFAIK lock files are created on demand so after boot the file wont be there

corest: That is a mount. Volume handles this like few rows below

corest: I didn’t see lock errors in kube-proxy

PR #166, opened 2018-10-23
Add metrics-server

corest:
Towards https://github.com/giantswarm/giantswarm/issues/4512


corest:
At the beginning we thought it is the replacement for cadvisor, which is deprecated. But now I don't think that is still true. We just need to make cadvisor standalone

corest:
- controllers and kubectl top read from /apis/metrics.k8s.io via metrics-server
- metrics pipelines, like Prometheus, sysdig, etc should read directly from the nodes' /metrics/cadvisor endpoint

we don't need a metrics-server in control-plane

PR #168, opened 2018-10-24
remove ansible installation

calvix:
we already do that in Dockerfile https://github.com/giantswarm/docker-terraform-and-stuff/blob/master/Dockerfile#L28

### DIFF
--- a/misc/e2e-azure.sh
+++ b/misc/e2e-azure.sh
@@ -64,8 +64,6 @@ exec_on(){
 }
 
 stage-preflight() {
-  apt update && apt -y install
-
   PROGS=( git terraform terraform-provider-ct terraform-provider-gotemplate az ansible-playbook ssh-keygen )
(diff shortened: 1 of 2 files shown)


misc/e2e-azure.sh
```diff
@@ -64,8 +64,6 @@ exec_on(){
 }
 
 stage-preflight() {
-  # TODO: move this to e2e-Dockerfile
-  pip install ansible -q -q -q --upgrade
   apt update && apt -y install
```
r7vme: oh what this `apt -y install` does?

calvix: I was also wondering what the hell is that?

r7vme: git blame shows

```
-  apt update && apt -y install ed
+  apt update && apt -y install
```

we are not using `ed` now, so we can remove it :D

PR #172, opened 2018-11-16
Use kube-proxy config from configmap

corest:
I also moved kubeconfigs into separate folder to not mess up the things with mounting separate files into same folder

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -2032,8 +2032,8 @@ storage:
 
             # create kube-proxy configmap
             while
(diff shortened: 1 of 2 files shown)


templates/master.yaml.tmpl
```diff
@@ -2013,6 +2030,33 @@ storage:
                 done
             done
 
+            # create kube-proxy configmap
```
corest: I changed this so that configmap is create or updated if exists

templates/master.yaml.tmpl
```diff
@@ -2013,6 +2030,33 @@ storage:
...
+                while
+                    /usr/bin/docker run --net=host --rm -v /srv:/srv $KUBECTL apply -f /srv/$manifest
+                    [ "$?" -ne "0" ]
+                do
+                    echo "failed to apply /srv/$manifest, retrying in 5 sec"
+                    sleep 5s
+                done
+            done
+
+            # restart ds to apply config from configmap
```
corest: restarting ds to force getting updated cm

PR #173, opened 2018-11-18
Setup aws vpn machine for china regions

calvix:
add VPN-instance module which is used for China regions to have private bastions

password for now in  envs same as token, in future we should move to encrypted file in installations

### DIFF
--- a/modules/aws/vpn_instance/vpn-instance.tf
+++ b/modules/aws/vpn_instance/vpn-instance.tf
@@ -34,6 +34,15 @@ resource "aws_instance" "vpn_instance" {
   }
 }
 
+resource "aws_eip" "vpn_eip" {
+  vpc      = true
+}
+
+resource "aws_eip_association" "vpn_eip" {
+  instance_id   = "${aws_instance.vpn_instance.id}"
+  allocation_id = "${aws_eip.vpn_eip.id}"
+}
+
 resource "aws_security_group" "vpn_instance" {
   name   = "${var.cluster_name}-vpn-instance"
   vpc_id = "${var.vpc_id}"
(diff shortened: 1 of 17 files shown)


modules/aws/vpn_instance/vpn-instance.tf
```diff
@@ -0,0 +1,134 @@
...
+resource "aws_route53_record" "vpn_instance" {
+  count   = "${var.route53_enabled ? 1 : 0}"
+  zone_id = "${var.dns_zone_id}"
+  name    = "vpn-instance${count.index + 1}"
+  type    = "A"
+
+  # Add "public_ip" or "private_ip" depending on "with_public_access" parameter.
+  records = ["element(aws_instance.vpn_instance.*.public_ip, count.index)"]
+  ttl     = "300"
+}
```
calvix: this probably doesn't make sense now, but can be useful for future when route53 is available in China

templates/vpn.yaml.tmpl
```diff
@@ -0,0 +1,140 @@
...
+          #!/bin/bash 
+          while ! curl --connect-timeout 5 http://169.254.169.254 >/dev/null; do sleep 1s; echo "waiting for vpc endpoint"; done;
+
+          PRIVATE_IP=$(curl http://169.254.169.254/latest/meta-data/local-ipv4) 
+          PUBLIC_IP=$(curl http://169.254.169.254/latest/meta-data/public-ipv4)
+
+          sed -i "s/__PRIVATE_IP__/${PRIVATE_IP}/g" /etc/strongswan/vpn-config.conf /etc/strongswan/ipsec.secrets /opt/bin/setup-iptables
+          sed -i "s/__PUBLIC_IP__/${PUBLIC_IP}/g" /etc/strongswan/vpn-config.conf /etc/strongswan/ipsec.secrets
+
+          echo "Files /etc/strongswan/vpn-config.conf /etc/strongswan/ipsec.secrets succesfully templated."
```
calvix: templating ipsec config and iptable rules, as this configs require the private and public ip info we don't have at time when terraform ignition is rendered

calvix:
some extra info:
https://github.com/giantswarm/adidas/issues/356#issuecomment-439686486

PR #175, opened 2018-11-21
remove test password for VPN

calvix:
I somehow forget to remove a test password from the variable files. 
Real password is now set via TF_VAR_xx  in `envs.sh` and is encrypted via opsctl.

### DIFF
--- a/platforms/aws/giantnetes/variables.tf
+++ b/platforms/aws/giantnetes/variables.tf
@@ -265,12 +265,10 @@ variable "external_ipsec_public_ip_1" {
   default     = "95.179.153.65"
 }
 
-## generation needs to be set somewhere else
-## for now just test password
+## vpn password
 variable "external_ipsec_password" {
   description = "shared password for ipsec connection"
   type        = "string"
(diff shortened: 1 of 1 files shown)


platforms/aws/giantnetes/variables.tf
```diff
@@ -265,12 +265,10 @@ variable "external_ipsec_public_ip_1" {
   default     = "95.179.153.65"
 }
 
-## generation needs to be set somewhere else
-## for now just test password
+## vpn password
 variable "external_ipsec_password" {
   description = "shared password for ipsec connection"
   type        = "string"
-  default     = "XHJYW67%45#0KJU(*56dvgyH(*$%$#TRi*G^^rxWEY"
```
tuommaki: So just to make sure: This is not used anywhere atm. (i.e. no revocation/change needed)?

calvix: Yeah, I used this when testing stuff,  but final installaion on giraffe has different password. You can easily check in installation repo by running `opsctl show secret -i terraform-secrets.yaml -k Terraform.VPNPassword`

PR #176, opened 2018-11-21
add-lifecycle-hook-to-asg

calvix:
add lifecycle hook to control plane ASG for future draining of workers before rolling nodes.

60 sec should be enough for now as we anyway don't have that much pods on nodes and draining is done in few sec, also if something breaks we will wait just extra 1 minutes for rolling

Maybe later we can increase to 120s


corest:
Is this still a thing? @calvix

calvix:
this was a part of graceful draining of AWS nodes, but i never had time to implement the rest of the details, closing for now

PR #182, opened 2018-11-30
double fs.inotify.max_user_watches

jgsqware:
Related to this PM
giantswarm/giantswarm#4777

This PR double the inotify watches

Signed-off-by: Julien Garcia Gonzalez <julien@giantswarm.io>


calvix:
> I wonder why there is no hardening.conf file on master?

maybe somebody forgot it, please create issue or PR :)

PR #186, opened 2018-12-13
add-worker-permissions-for-cloudwatch-logs

calvix:
add perm for forwarding logs to cloudwatch to worker IAM

I think it makes sense to have this by default

towards https://github.com/giantswarm/giantswarm/issues/4830


calvix:
RFR in order to close the issue, for now, we can iterate later if necessary.

PR #190, opened 2019-01-07
multi master AWS

calvix:
add support for multi-master AWS CP

changes:
* 3 masters, each master in  different AZ
* reintroduced etcd ELB, which is now loadblancing between 3 etcds
* etcd cluster is created via definiging 3 etcd peers in conf, no discovery settings
* halved worker and elb subnets in order to have 3(+1 unused) subnets for third zone
* each master have sapareate ignition with unique master_id (1, 2 or 3) 
* with 3 zones we now have 3 EIP and 3 NAT GW
* masters have `ignore_changes = [*]` set, so we dont accidentaly dont roll all three masters, if ts necessery to roll masters you should do it one by one with tainting them 

for migration i have text-guide

umbrella issue https://github.com/giantswarm/giantswarm/issues/5062


calvix:
not yet for merging but wanted to put the changes somewhere

calvix:
damn, China  Beijing has only 2 AZ :(

calvix:
forcing use of `element` interpolation function to access items in array will give us nice workaround for zones with lower than 3 AZs

see definition:
```
element(list, index) - Returns a single element from a list at the given index. If the index is greater than the number of elements, this function will wrap using a standard mod algorithm. This function only works on flat lists. 
```

so in the case of 2 AZs the third resource (worker subnet, bastion subnet and others) will fall into AZ 1, it wont be true HA but we will avoid some weird hacking

calvix:
last step is adding proper migration guide

calvix:
@giantswarm/sig-infrastructure if you wanna take a look as this code si final

modules/aws/master/master-elb.tf
```diff
@@ -24,18 +24,44 @@ resource "aws_elb" "master" {
...
-      "Name", "${var.cluster_name}-master"
+      "Name", "${var.cluster_name}-master-api"
     )
   )}"
 }
 
-resource "aws_elb_attachment" "master" {
-  elb      = "${aws_elb.master.id}"
-  instance = "${aws_instance.master.id}"
+resource "aws_elb" "master_etcd" {
```
mazzy89: Definitely, I'm missing here some previous discussion about the architecture, but why etcd must be load balanced and reachable from the external?

modules/aws/master/master.tf
```diff
@@ -18,69 +15,85 @@ resource "aws_instance" "master" {
...
 
   tags = "${merge(
     local.common_tags,
     map(
       "Name", "${var.cluster_name}-master${count.index}"
     )
   )}"
+
+  # we ignore changes, to avoid rolling all masters at once
+  # update is done via tainting masters
```
mazzy89: This could be an improvement for the future may be a discussion for the SIG Infra.

We could put Masters under a separate ASG. So 3 AZs = 3 ASGs. When a new Launch Configuration is detected (a change in the user data) automatically a new ASG is created in TF. Then with a separate script we:

- drain each master
- terminate the instance. Automatically a new instance will be created (since we run in ASG) with a new LaunchConfiguration.

The process is automated and less error prone. WDTY?

modules/aws/vpc/vpc-subnet-worker.tf
```diff
@@ -24,6 +24,19 @@ resource "aws_subnet" "worker_1" {
   )}"
 }
 
+resource "aws_subnet" "worker_2" {
```
mazzy89: These resources here looks pretty redundant. Would be possible to create an unique resource `aws_subnet` and leverage interpolation?

modules/aws/vpc/vpc-subnet-worker.tf
```diff
@@ -33,3 +46,8 @@ resource "aws_route_table_association" "worker_1" {
   subnet_id      = "${aws_subnet.worker_1.id}"
   route_table_id = "${aws_route_table.cluster_vpc_private_1.id}"
 }
+
+resource "aws_route_table_association" "worker_2" {
```
mazzy89: Same here as said about `aws_subnet`

modules/aws/vpc/vpc-subnet-worker.tf
ferrandinand: Not sure if you can interpolate inside the resource name.

You can create modules or one resource with count.
Anyway it is always a good idea to reduce redundant code.

mazzy89: > Not sure if you can interpolate inside the resource name.

No you can't. Terraform resource name should not know anything about the count of the resource. we should call it simply `worker` and then interpolate with `count` as you said

modules/aws/master/master-elb.tf
calvix: As you can see few lines below your comment there is `  internal = true` so this is internal loadblancer so its not reachable from external sources

mazzy89: Sorry. Right. I missed that. So it is internal so not reachable. Another question is: for which use cases it is used?

modules/aws/master/master.tf
calvix: Yes, we already thought about this but the main problem with ASG is that if your instance dies, all EBS are gone, and this is a problematic for etcd cluster as its very easy to break cluster when some of your nodes are dead.

 with simple ec2 if the master goes down we still have ebs and rerunning terraform will start a new instance with the same data and we are 100% sure we wont break etcd cluster.

AWS quite often schedule hardware for maintenance or have other events that can cause your instance to be deleted which would in our case means that etcd data on one machine is lost.

In general, it would be nice to have  ASG only, but having a very reliable etcd cluster under ASG is very complex. There is a nice blog post about it and it recommends 9 ASGs and 9 nodes in the cluster ( 3 per each AZ).

modules/aws/vpc/vpc-subnet-worker.tf
calvix: Yep, good suggestion, I think we should be able to do that,  but we would need to move subnet values into array

modules/aws/master/master-elb.tf
calvix: calico is connecting to etcd so its mostly for calico

modules/aws/master/master.tf
```diff
@@ -18,69 +15,85 @@ resource "aws_instance" "master" {
...
+  lifecycle {
+    ignore_changes = ["*"]
```
ferrandinand: If we are planning to taint and then apply to protect the rollout maybe we can include here  create_before_destroy = true, to avoid the lose of one master if the creation fails.

Extra point would be create a local-exec script in this resource to check the availability of the resource before destroying the old one. If the resource is created but the health script fail won't be destroyed either.

modules/aws/master/master-elb.tf
mazzy89: Right calico need etcd endpoints in the configuration. This might be another proposal for SIG Infra. What about to remove the ELB from the equation in the future and pass to calico the DNS name of each master to the etcd port (2379 if I'm not wrong). of course, this introduces the problem to keep update the DNS record with the IP address of the master. But this should be probably easily solvable with the introduction of a bit of automation which come with a tool like external-dns. would be possible or do you find any big constraints? The reason for this proposal is because I'm always happy when I'm able to remove from the equation piece of external infra like ELB which participate to increase the bill and have a more neat infra.

modules/aws/master/master.tf
mazzy89: I see that's true. Thank you for the clarification. But would be an amazing improvement (and maybe not hard to implement) something like at the bootstrap of the master node look for any suitable etcd volume (if exists) to mount. Otherwise create a new one. And with this would not matter anymore if we run an EC2 instance inside or outside an ASG.

modules/aws/master/master.tf
mazzy89: 👍 Good level of automation. I like that.

calvix: That's a good idea, But I don't want to add more complexity to this story, this would be a nice fit for a separate issue we can address later, can you create an issue to sig-infra @ferrandinand

ferrandinand: ok, let´s do that.

mazzy89: Agree to address this in a different PR.

docs/migrate-to-multi-master-guide-aws.md
```diff
@@ -0,0 +1,112 @@
...
+ETCDCTL_API=3 etcdctl snapshot restore  --data-dir="/tmp/etcd"  db-snapshot
+sudo cp -r /tmp/etcd/member /var/lib/etcd/
+sudo rm -rf /tmp/etcd/
+sudo systemctl start etcd3
+```
+
+
+Export envs to use etcdctl againts cluster:
+```
+export ETCDCTL_CERT_FILE=/etc/kubernetes/ssl/etcd/client-crt.pem
```
corest: in latest master this configuration already done with alias
https://github.com/giantswarm/giantnetes-terraform/blob/master/templates/master.yaml.tmpl#L2686-L2691

corest: That's now is tricky. Alias references lb and endpoint will be overridden even if you export domain with `etcd1` dns

corest: So for migration it is make sense to add mention about cleaning up alias with `alias etcdctl=` in sudo mode

platforms/aws/giantnetes/variables.tf
```diff
@@ -199,49 +199,27 @@ variable "k8s_api_ip" {
...
-
-variable "subnet_worker_1" {
-  description = "CIDR for worker network 1."
-  default     = "10.0.5.64/26"
-}
-
-variable "subnet_worker_2" {
-  description = "CIDR for worker network 2."
-  default     = "10.0.5.128/26"
+variable "subnet_worker" {
```
mazzy89: 🎉 well done. just a small note. would be better have these variable in plural like `subnet_workers`?

calvix: Yeah naming is bad, will try to figure out something, so far I like that same thing start with the same word so when ordering  they are  at the same place:
subnets_worker
subnets_bastion
subnets_elb
```

File: modules/aws/master/master.tf (hunk 2 above)
Comment (calvix) [reply modules/aws/master/master.tf]: You put the idea into a new issue and put it into sig-infra, if you have and idea how to do this reliably than sure let's do it.

File: modules/aws/master/master-elb.tf (hunk 1 above)
Comment (calvix) [reply modules/aws/master/master-elb.tf]: That might work too, lets create issue fro that and we can easily adress later.

File: docs/migrate-to-multi-master-guide-aws.md (hunk 6 above)
Comment (calvix) [reply docs/migrate-to-multi-master-guide-aws.md]: `{{ .ETCDDomainName }}` on master is actually etcd1 or etcd2 or etcd3 depending on which master it is so we are fine

File: modules/aws/master/master-elb.tf (hunk 1 above)
Comment (mazzy89) [reply modules/aws/master/master-elb.tf]: Going to do that ;-)

File: modules/aws/master/master.tf (hunk 2 above)
Comment (mazzy89) [reply modules/aws/master/master.tf]: Going to do that ;-) thank you for the follow up on this.

File: docs/migrate-to-multi-master-guide-aws.md (hunk 6 above)
Comment (corest) [reply docs/migrate-to-multi-master-guide-aws.md]: oh, nice then

File: docs/migrate-to-multi-master-guide-aws.md [hunk 8]
```diff
@@ -0,0 +1,138 @@
...
+```
+after deletion go back to latest master
+
+### update terraform values
+There is change in subnet naming and structure. Subnets are agregated into lists and names are also changed: 
+```
+subnets_bastion = ["10.0.1.0/25", "10.0.1.128/25"]
+subnets_elb = ["10.0.2.0/26", "10.0.2.64/26", "10.0.2.128/26"]
+subnets_vault = ["10.0.3.0/25"]
+10.0.220.48/28subnets_worker = ["10.0.5.0/26", "10.0.5.64/26", "10.0.5.128/26"]
```
corest: typo here?

docs/migrate-to-multi-master-guide-aws.md
```diff
@@ -0,0 +1,138 @@
...
+```
+## Recover rest of cluster
+* Restart kubelets on all 3 masters to ensure everyhting is up and running.
+* You might need to restart all workers so they properly register to k8s-api.
+* On one of masters restart `k8s-addons` to esnure latest version is applied.
+
+## Recover tenant cluster peering
+If there are any cluster created with `aws-operator`, then the CloudFOrmation stack that is configuring VPC peergin needs to be deleted  in order to reconfigure proper connection between VPC.
+
+The CLoudFormation stack should look like this: `cluster-{cluster_id}-host-main`. After deletion, `aws-operator` will recreate it with proper values.
```
corest: `CLoudFormation` -> `CloudFormation`

PR #191, opened 2019-01-10
Disable insecure port and anon auth

r7vme:
related: https://github.com/giantswarm/giantswarm/issues/2984

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -2092,25 +2092,30 @@ storage:
       contents:
         inline: |
             #!/bin/bash
-            # kubectl 1.12.2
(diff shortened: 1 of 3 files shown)


templates/master.yaml.tmpl
```diff
@@ -2370,8 +2370,10 @@ storage:
                 - /hyperkube
                 - apiserver
                 - --allow-privileged=true
+                - --anonymous-auth=false
                 - --kubelet-https=true
                 - --secure-port=443
+                - --insecure-port=0
```
kopiczko: Usually 0 is a randomly used port. Isn't that the case?

calvix: No, in this case 0 is disabled

PR #194, opened 2019-01-23
Mount locally dockershim

mazzy89:
Running kubelet as a container does not expose the `dockershim` socket to the host automatically. By exposing it we can access the common kubernetes CR-I and have enormous benefits in debugging our applications remotely running in the control plane. 

For instance a series of tools out there like https://github.com/solo-io/kubesquash use the common CR interface to operate against the docker API.

Doing that we can also build tools that can operate remotely.

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -3076,6 +3076,7 @@ systemd:
       -v /run/calico/:/run/calico/:rw \
       -v /run/docker/:/run/docker/:rw \
       -v /run/docker.sock:/run/docker.sock:rw \
+      -v /run/dockershim:/run/dockershim:rw \
       -v /var/log:/var/log:rw \
       -v /usr/lib/os-release:/etc/os-release \
       -v /usr/share/ca-certificates/:/etc/ssl/certs \
@@ -3096,6 +3097,7 @@ systemd:
       /hyperkube kubelet \
       --config=/etc/kubernetes/config/kubelet.yaml \
       --node-ip=${DEFAULT_IPV4} \
+      --container-runtime-endpoint=/var/run/dockershim/dockershim.sock \
       --containerized \
       --enable-server \
       --logtostderr=true \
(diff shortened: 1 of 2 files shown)


templates/master.yaml.tmpl
```diff
@@ -3096,6 +3097,7 @@ systemd:
       /hyperkube kubelet \
       --config=/etc/kubernetes/config/kubelet.yaml \
       --node-ip=${DEFAULT_IPV4} \
+      --container-runtime-endpoint=/var/run/dockershim/dockershim.sock \
```
tuommaki: I understood that default path for this would be `/var/run/dockershim.sock` - is there some reason to put it under own directory?

I think that e.g. use of `kubesquash` would be easier when you don't need to know and specify the location of this socket: https://github.com/solo-io/kubesquash/blob/master/cmd/kubesquash/main.go#L44 WDYT?

mazzy89: Completely agree with that However I've already done the try but kubelet doesn't start and get this error:

```
`Jan 23 22:17:21 ip-10-0-5-235 sh[16181]: F0123 22:17:21.188498   16208 server.go:261] failed to run Kubelet: failed to create kubelet: failed to listen on "unix:///var/run/dockershim.sock": failed to unlink socket file "/var/run/dockershim.sock": is a directory`
```

this because dockershim.sock does not exist on the host, docker thought you intention is to mount it as a directory
so it created an empty directory there instead

That is the reason for this choice

tuommaki: Oh, wow! Makes sense. Would be perhaps nice to check later if there's a way to get this fixed, but for now I think this is a good workaround.

tuommaki:
Also wait that one of the SREs approves this.

mazzy89:
> Make sure that CI is green, (probably you need to retry the azure job)

Yeah ok the Azure job looks a bit flaky.

> Is there any security issue or danger when we expose the container-runtime? What can it be used for?

No security issues at all. In any kubernetes installation actually `dockershim.sock` is by default in the machine like `docker.sock` because kubelet does not run as we do in a container. You have to think to `dockershim.sock` like if you think to to the `docker.sock`

> In general, exposing debug things is not something we want to do on production. Isn't there any other way how you could debug aws-operator?

We are not exposing any debug tools. Simply the Container Runtime Interface of docker which correspond to `dockershim`.

PR #196, opened 2019-02-05
Add calicoctl/etcdctl/crictl tools for master

corest:
Towards https://github.com/giantswarm/giantswarm/issues/5216

I don't want to use containers for those debug tools as that adds an additional point of failure in the debug process. But that's just me. Do you think we need containers for those utils?

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -2094,15 +2094,6 @@ storage:
             #!/bin/bash
             set -eu
 
-            # download etcdctl
-            ETCD_VERSION=v0.4.5
-            wget https://github.com/etcd-io/etcd/releases/download/${ETCD_VERSION}/etcd-${ETCD_VERSION}-linux-amd64.tar.gz
-            tar xvf etcd-${ETCD_VERSION}-linux-amd64.tar.gz
-            mv etcd-${ETCD_VERSION}-linux-amd64/etcdctl /opt/bin/etcdctl
-            chmod +x /opt/bin/etcdctl
-            rm etcd-${ETCD_VERSION}-linux-amd64.tar.gz
-            rm etcd-${ETCD_VERSION}-linux-amd64 -Rf
-
             # download calicoctl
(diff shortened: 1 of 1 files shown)


templates/master.yaml.tmpl
```diff
@@ -2086,6 +2086,58 @@ storage:
...
+            set -eu
+
+            # download etcdctl
+            ETCD_VERSION=v0.4.5
+            wget https://github.com/etcd-io/etcd/releases/download/${ETCD_VERSION}/etcd-${ETCD_VERSION}-linux-amd64.tar.gz
+            tar xvf etcd-${ETCD_VERSION}-linux-amd64.tar.gz
+            mv etcd-${ETCD_VERSION}-linux-amd64/etcdctl /opt/bin/etcdctl
+            chmod +x /opt/bin/etcdctl
+            rm etcd-${ETCD_VERSION}-linux-amd64.tar.gz
+            rm etcd-${ETCD_VERSION}-linux-amd64 -Rf
```
calvix: etcdctl is part of CoreOS distribution so I am not so sure we really need to download yet another etcdctl

templates/master.yaml.tmpl
```diff
@@ -3228,6 +3291,17 @@ systemd:
...
+    enabled: true
+    contents: |
+      [Unit]
+      Description=Calicoctl, etcdctl and crictl tools
+      After=network.target
+      [Service]
+      Type=oneshot
+      ExecStart=/opt/install-debug-tools
+      [Install]
+      WantedBy=multi-user.target
```
calvix: do we need to install it automatically? from a security perspective, it  would be best to just have an install file that we can run on demand, but that just opinion, not a blocker

templates/master.yaml.tmpl
corest: true

templates/master.yaml.tmpl
corest: That's an additional context you should know. But also no strong opinion on this

PR #200, opened 2019-02-11
use-multi-etcd-endpoints-instead-of-ELB

calvix:
remove extra elb we don't need and use multi etcd endpoint

as @mazzy89 suggested, using ELB  for etcd is not necessary as we can add multiple endpoints for etcd in calico conf, we already do the same for on-prem so we are just aligning here


mazzy89:
Good one. 

I have one question. Who does ensure that etcd DNS records are in sync?


PS. Just one quick note @calvix also recommended to go with this solution here https://github.com/giantswarm/giantswarm/issues/5317#issuecomment-461936247

calvix:
switch to use calico-typha for whole networking is out of scope of this PR, that is a big change we should properly test  and prepare for,  will probably close this PR for now

corest:
@calvix close? PR reminder keep sending me this

calvix:
I will give it a quick spin and see if we can easily merge it otherwise I will close it.

calvix:
all green, nice

kopiczko:
Are you confident about DNS propagation? This is the reason we switched back to ELB in TC.

kopiczko:
Unless it doesn't go trough Route53.

calvix:
> Are you confident about DNS propagation? This is the reason we switched back to ELB in TC.

Not sure about the TC but in CP we have TTL for etcd records 30 sec which is low enough to work this nicely. At least that's what I saw in tests. Let's see how it behaves in real envs.

PR #202, opened 2019-02-13
azure-multi-master

calvix:
azure multiMaster

lots of things are simplified as the whole module for azure was written  in regards to multiple masters and i already did few modifications to code in aws multimaster

* adding multi-master support ( similar to aws we can fallback to single master)
* updating e2e scripts
* update circleci config fro multi and single master e2e tests
* docs for migration

towards https://github.com/giantswarm/giantswarm/issues/5061


platforms/azure/giantnetes/variables.tf
```diff
@@ -204,41 +210,6 @@ variable "k8s_api_ip" {
...
-variable "subnet_worker_0" {
-  description = "CIDR for worker network 0."
-  default     = "10.0.5.0/25"
-}
-
-variable "subnet_worker_1" {
-  description = "CIDR for worker network 1."
-  default     = "10.0.5.128/25"
-}
-
```
calvix: we don't  use these variables

docs/migrate-etcd-to-multi-cluster.md
```diff
@@ -0,0 +1,75 @@
...
+sudo vim /etc/systemd/system/etcd3.service
+sudo systemctl daemon-reload
+systemctl start etcd3
+```
+
+Now you should have 3 healthy nodes in etcd cluster, to ensure run:
+```
+sudo -E etcdctl cluster-health
+```
+
```
calvix: I have moved the etcd migration to separate file as its same for both cloud providers

PR #203, opened 2019-02-14
Terraform semver for provider and terraform blocks

mazzy89:
From https://github.com/giantswarm/giantswarm/issues/5215

Summary:

- Remove `provider.tf` and `backend.tf` and moved them in the provider `main.tf` and added there the fixed version for both.
- Renamed `envs.sh` to `bootstrap.sh` and moved `terraform init`
- Update docs

Not tested yet. I need to do a full run of the entire setup to verify that everything still works and nothing is broken.


ferrandinand:
Looks fine to me but just a question, what is the reason to not use tfvars file instead a sh to export.
Is because TF_VARS takes precedence and we don´t want to be overwritten in any case?

Mixed feeling about that but LGTM.

mazzy89:
that's a good observation @ferrandinand. I haven't thought about that honestly because I did not want to diverge too much from the current setup but it would be nice to have the input of the seniors here around.

calvix:
what do you mean by `tfvars` ? Can you elaborate more?

In general, we have many variables in each provider that have some defaults (https://github.com/giantswarm/giantnetes-terraform/blob/master/platforms/aws/giantnetes/variables.tf ), but we need to change them for some installations ( like subnets, regions,....), also shell script can inject easily encrypted secrets via executing `opsctl`

If you think we can somehow do that better than shell than let's do that!

mazzy89:
You can create a file named for instance `terraform.tfvars` put in there vars and call it from each `terraform` command like `terraform init -var-file=terraform.tfvars` and `terraform plan -var-file=terraform.tfvars` and so on.

Personally not opinionated about this. Don't have any strong arguments for either env vars and tfvars. If @ferrandinand do you feel that can help the process making it less error prone then let's go with that.

calvix:
@mazzy89 btw e2e failed because you did not run `terraform fmt` on terraform files

mazzy89:
Yes. thank you @calvix  at the moment I'm trying to test. going to fix with the upcoming commits.

ferrandinand:
Yes, the way we use it is totally fine .

The differences between using TF_VARS and a variable file are subtle 
The first one is in the precedence
1. Env vars
2. terraform.tfvars
3. var-file= depending on the order (left to right order)

So if we want to override .sh TF_VARS we won´t be able.

Also It is a good practice, if an infrastructure code grow, separate in different tfstate files the code.
We could run terraform multiple times like terraform apply -var-file=xxx && terraform apply -var-file=yyy without setting an unsetting env vars.

What it comes to secrets, they injecting is easier in the way you do but can be encryted in a separated tfvars files then uncrypted with permissions and executed like this  terraform plan -var-file=avatar.tfvar -var-file=avartar-secrets.tfvars similiar what is done for draughman

Maybe it is not the PR to discuss about it just another point of view (not necessary better) based on my use of terraform

mazzy89:
Nice follow up here. Azure provider version 1.22.0 is broken. 1.22.1 fixed the issue. Bumped and it got solved 😃

mazzy89:
> if an infrastructure code grows, separate in different tfstate files the code.

Completely agree with that ❤️

> What it comes to secrets, they injecting is easier in the way you do but can be encrypted in a separated tfvars files then decrypted with permissions and executed like this terraform plan -var-file=avatar.tfvar -var-file=avartar-secrets.tfvars similar what is done for draughman

Agree too here.

It is worths to talk about this in the next SIG Infra @ferrandinand

corest:
Is this branched from latest master? I see it mixed with multi-master migration and it is hard to see what is the actual change

mazzy89:
It has been rebased 1hour ago with the master. I'm going to rebase it again to solve the conflicts

mazzy89:
Rebased and running again the updated e2e tests

mazzy89:
All tests green. Fully tested.any last things to adreess before merging @calvix @corest?

mazzy89:
We hung with Vol offline to discuss this PR and we agreed to move out again `terraform init` from the `bootstrap.sh`, rename `bootstrap.sh` back to `envs.sh` and run `terraform init` along with backend params. 

Simply `terraform init` did not look nice inside `bootstrap.sh` because it was along with envs variable and a bit hidden. He would prefer overall to call `init` explicity.

calvix:
Why? I would overall prefer something that does require remember some commands or copy them always from some guide, I would like to have automated it so I don't have to care.

mazzy89:
Yes I agree with you @calvix. I also prefer having fewer commands to type and remember. @corest anything to add about this?

corest:
I don't like the approach of mixing data and init process within the same file. Having `envs.sh` means we store data and inject it elsewhere. It is clear flow: load data -> use it somewhere. Now we are mixing stuff. If this mixing is a requirement for validation of the terraform version - I'll agree on that. I agree we need this, but don't agree on the change. Anyway, 2 votes > 1 vote. Feel free to merge

mazzy89:
I completely see your point Vol and  I could partially agree but I think that in comparison of the gain we have by delivering this improvement, we could live for a while with this approach. Consider that I'm saying this because already as next we could Implement the improvement mentioned by Ferran that is to use terraform tfvars which is exactly the reason why it exists that flag that is load data which is what we are talking in here. So I'd go with that and then take as next the tfvars load which address all our concerns.

corest:
We have a quorum on approval. That's enough to merge. So let's move here

PR #207, opened 2019-02-20
define-backend-type

calvix:
without the backend definition terraform does not know which backend type use


mazzy89:
Yes, good catch. I've missed it.

PR #208, opened 2019-02-25
Add giantswarm-critical priority class

corest:
Align with tenant clusters so we can add priority class `giantswarm-critical` for CP components (we can't reuse system priorityclasses as those works only for `kube-system` ns)

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -1152,6 +1152,18 @@ storage:
             name: calico-node
             namespace: kube-system
           {{- end }}
+    - path: /srv/priority_classes.yaml
+      filesystem: root
+      mode: 0644
+      contents:
+        inline: |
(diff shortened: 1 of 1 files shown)


templates/master.yaml.tmpl
```diff
@@ -2166,6 +2178,17 @@ storage:
...
             done
 
+            # apply priority classes
+            PRIORITY_CLASSES_FILE="priority_classes.yaml"
+
+            while
+                $KUBECTL apply -f /srv/${PRIORITY_CLASSES_FILE}
+                [ "$?" -ne "0" ]
+            do
+                echo "failed to apply /srv/${PRIORITY_CLASSES_FILE}, retrying in 5 sec"
```
mazzy89: Just to understand here. Why here would `kubectl` fail?

corest: Dunno. Flapping api, flapping network, etc

PR #211, opened 2019-02-27
Remove build dir

calvix:
adjust docs and e2e to remove build dir


calvix:
finally this works :P

calvix:
We need to keep the sed lien for removing the backend line, in the original e2e with build dir and `backend.tf` files we just never created proper backend configuration. Now the behavior and content are the same.

mazzy89:
Yes the `sed` is to let to use `local` as backed rather than `s3` or `azure`

calvix:
RFR

calvix:
e2e failing because azure is deleting resources for 2 hours+, but otherwise all nodes are ready, merging

PR #212, opened 2019-02-27
KMS setup for vault auto-unseal in AWS

corest:

### DIFF
--- a/modules/aws/vault/vault-iam.tf
+++ b/modules/aws/vault/vault-iam.tf
@@ -32,14 +32,14 @@ data "aws_iam_policy_document" "vault-kms-unseal" {
     actions = [
       "kms:Encrypt",
       "kms:Decrypt",
-      "kms:DescribeKey"
+      "kms:DescribeKey",
     ]
   }
 }
 
 resource "aws_iam_role_policy" "vault-ksm-unseal" {
-  name = "${var.cluster_name}-vault-kms-unseal"
-  role = "${aws_iam_role.vault.id}"
+  name   = "${var.cluster_name}-vault-kms-unseal"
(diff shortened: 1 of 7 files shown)


teemow:
the changes lgtm. but can you explain a bit this will look in the end? is kms used within the same account the installation is in or a different?

modules/aws/vault/vault-iam.tf
```diff
@@ -0,0 +1,58 @@
...
+
+    actions = [
+      "kms:Encrypt",
+      "kms:Decrypt",
+      "kms:DescribeKey"
+    ]
+  }
+}
+
+resource "aws_iam_role_policy" "vault-ksm-unseal" {
```
mazzy89: Typo

modules/aws/vault/vault-iam.tf
```diff
@@ -0,0 +1,58 @@
...
+  name = "${var.cluster_name}-vault"
+  role = "${aws_iam_role.vault.name}"
+
+  lifecycle {
+    create_before_destroy = true
+  }
+
+  # Sleep a little to wait the IAM profile to be ready -
+  provisioner "local-exec" {
+    command = "sleep 10"
```
mazzy89: Look very hackish. Which is the reason for that? The reference is pretty explicit in the `aws_instance.vault` in the line

```hcl
  iam_instance_profile = "${aws_iam_instance_profile.vault.id}"
```

so Terraform will care about that.

corest:
@teemow it is used within the same account, where CP is installed.

modules/aws/vault/vault-iam.tf
corest: copy past from worker

PR #213, opened 2019-03-02
Add key vault for auto-unseal

corest:

### DIFF
--- a/modules/azure/vault/vault.tf
+++ b/modules/azure/vault/vault.tf
@@ -29,7 +29,7 @@ resource "azurerm_virtual_machine" "vault" {
 
   identity = {
     type         = "UserAssigned"
-    identity_ids = ["${azurerm_user_assigned_identity.vault.id}"]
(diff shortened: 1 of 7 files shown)


modules/azure/vault/vault.tf
```diff
@@ -14,6 +14,11 @@ resource "azurerm_virtual_machine" "vault" {
   network_interface_ids = ["${var.network_interface_ids[0]}"]
   vm_size               = "${var.vm_size}"
 
+  identity =
+  {
+    type = "SystemAssigned"
+  }
```
kopiczko: Will it work on german cloud (if we still support it)?

corest: Nope. But I'll add conditionals for aws china and azure german/stack in separate PR

corest:
Important docs update on requirements of the `terraform` AD group added. @giantswarm/sig-infrastructure ptal

corest:
Merging as destroy takes ages on azure

PR #214, opened 2019-03-05
update cores and hyperkube

calvix:
remove CoreOS alpha channel and upgrade k8s with CVE fix

### DIFF
--- a/platforms/aws/giantnetes/variables.tf
+++ b/platforms/aws/giantnetes/variables.tf
@@ -99,10 +99,9 @@ variable "worker_instance" {
 }
 
 ### Container Linux ###
-## temporary until 1995.x will be in stable, as older version have disk timeout issues
 variable "container_linux_channel" {
   description = "Container linux channel (e.g. stable, beta, alpha)."
-  default     = "alpha"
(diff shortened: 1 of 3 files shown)


platforms/aws/giantnetes/variables.tf
```diff
@@ -99,10 +99,9 @@ variable "worker_instance" {
 }
 
 ### Container Linux ###
-## temporary until 1995.x will be in stable, as older version have disk timeout issues
 variable "container_linux_channel" {
   description = "Container linux channel (e.g. stable, beta, alpha)."
-  default     = "alpha"
+  default     = "stable"
```
mazzy89: Which will be the version that will be deployed?

mazzy89: Yep make sense now. you replied with that https://github.com/giantswarm/installations/pull/670/commits/ce532fd5fe6769b4be8f8f025d224806bfe86473

calvix: in general, the latest is deployed but we keep the version per installation in the `bootstrap.sh` file

PR #215, opened 2019-03-06
use-proper-hostname

calvix:
towards https://github.com/giantswarm/giantswarm/issues/5460

also, adjust settings for CP to be able to pass upstream conformance test

similar to this https://github.com/giantswarm/aws-operator/pull/1434 or tenants

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -3172,6 +3172,22 @@ systemd:
 
       [Install]
       WantedBy=multi-user.target
+  - name: set-hostname.service
+    enabled: true
+    contents: |
+      [Unit]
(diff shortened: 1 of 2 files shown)


mazzy89:
It makes senses but we could simply achieve this in a slighlty different way. `kuebelet` can be configured with a flag `--hostname-override` which can be set to the hostname returned by the EC2 metadata. In this way we would leave untouched the hostname of the AWS instances. From my experience I always had has hostname in EC2 instances something like `ip-xxx.xx.xxx.xx`. Changing the hostname does not have any effect?

calvix:
Make sense, but i wonder if we should keep stuff aligned with tenants  in this case.

PR #217, opened 2019-03-06
Setup auto-unseal for supported envs only

corest:


corest:
multi-master just fails on destroy

corest:
@giantswarm/sig-infrastructure ptal pls

PR #218, opened 2019-03-08
Get back kube-proxy to iptables mode

ferrandinand:
Towards:https://github.com/giantswarm/giantswarm/issues/5542

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -1376,7 +1376,7 @@ storage:
           clientConnection:
             kubeconfig: /etc/kubernetes/kubeconfig/kube-proxy.yaml
           kind: KubeProxyConfiguration
-          mode: ipvs
+          mode: iptables
           resourceContainer: /kube-proxy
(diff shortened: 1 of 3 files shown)


ferrandinand:
Tested on gauss -- OK

templates/master.yaml.tmpl
```diff
@@ -1423,7 +1423,7 @@ storage:
                   command:
                   - /hyperkube
                   - proxy
-                  - --proxy-mode=ipvs
+                  - --proxy-mode=iptables
```
TheoBrigitte: Is it possible to remove this flag ? and therefore only having the proxy-mode configured in one place (`kube-proxy.yaml`).

ferrandinand:
There is an upstream issue related with s3 destroy.
https://github.com/terraform-providers/terraform-provider-aws/issues/423

And hitting some vault resources

templates/master.yaml.tmpl
ferrandinand: I guess this will be temporary until ipvs gets stable, I guess makes sense to keep that in order to make easier future change to ipvs.

PR #220, opened 2019-03-09
WIP: Terraform 12 

corest:
Towards https://github.com/giantswarm/giantswarm/issues/3757
While testing Azure Stack, I did following changes to make terraform compatible with version 12


corest:
This PR is here just to save some time in the future

tuommaki:
> This PR is here just to save some time in the future

i.e. do you want review here or just keep this as working ground?

corest:
@tuommaki That's now ready for review

corest:
`AWS` is fine
`Azure` is completely unreliable (e.g. `>30min` to run vault node)

corest:
Tested manually on azure - worked fine

modules/aws/vpn/vpn.tf
```diff
@@ -16,20 +16,28 @@ resource "aws_vpn_gateway" "vpn_gw" {
...
     "giantswarm.io/installation" = "${var.aws_cluster_name}"
   }
 }
 
+data "aws_vpn_gateway" "vpn_gw" {
+  # Workaround for a bug that claimed to be addressed by 0.12 version.
+  # https://github.com/hashicorp/terraform/issues/12570
+  count = "${var.aws_customer_gateway_id_0 == "" ? 0 : 1}"
+
+  id = "${aws_vpn_gateway.vpn_gw.*.id[count.index]}"
```
tuommaki: What's the thing here with `*` and having `[count.index]` in the end? Could this be like following:
```suggestion
  id = "${aws_vpn_gateway.vpn_gw.[count.index].id}"
```

...below in untouched part there seems to be a second occurance of this.

modules/aws/vpn/vpn.tf
```diff
@@ -65,12 +73,12 @@ resource "aws_route" "vpc_route_0" {
...
-  gateway_id             = "${aws_vpn_gateway.vpn_gw.id}"
+  gateway_id             = "${aws_vpn_gateway.vpn_gw[0].id}"
 }
 
 resource "aws_route" "vpc_route_1" {
   count                  = "${var.aws_customer_gateway_id_1 == "" ? 0 : 2}"
   route_table_id         = "${var.aws_private_route_table_ids[count.index]}"
   destination_cidr_block = "${var.aws_external_ipsec_subnet_1}"
-  gateway_id             = "${aws_vpn_gateway.vpn_gw.id}"
+  gateway_id             = "${aws_vpn_gateway.vpn_gw[0].id}"
```
tuommaki: `0` or `1`?
```suggestion
  gateway_id             = "${aws_vpn_gateway.vpn_gw[1].id}"
```

modules/aws/vpn/vpn.tf
corest: This block should be dropped completely. It is not in use

modules/aws/vpn/vpn.tf
corest: No, it should be `0` here. `vpn_gw` has only one resource. But new terraform explicitly asks for list references. So if vpn vpc route should be created - it will always have same gateway.

PR #229, opened 2019-03-29
Fix kubelet labels

corest:
Towards https://github.com/giantswarm/giantswarm/issues/5560

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -3284,7 +3284,7 @@ systemd:
       --register-node=true \
       --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
       --kubeconfig=/etc/kubernetes/kubeconfig/kubelet.yaml \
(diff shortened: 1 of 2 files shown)


templates/master.yaml.tmpl
```diff
@@ -3284,7 +3284,7 @@ systemd:
       --register-node=true \
       --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
       --kubeconfig=/etc/kubernetes/kubeconfig/kubelet.yaml \
-      --node-labels="node-role.kubernetes.io/master,role=master,ip=${DEFAULT_IPV4},master-id={{.MasterID}}" \
```
tuommaki: Before merging / propagating this, I think we need to take care of current uses of `role=master`. I for example just created one node selector based on this for tenant last week (and yes, I understand this is for CP, but consistency).

calvix:
we can probably close this as fixing kubelet labels was part of previous k8s release

PR #233, opened 2019-04-08
Vault logs separate volume

corest:
Towards https://github.com/giantswarm/giantswarm/issues/4018

Thoughts?


modules/aws/vault/vault.tf
```diff
@@ -32,7 +42,7 @@ resource "aws_instance" "vault" {
     volume_size = "${var.volume_size_root}"
   }
 
-  user_data = "${var.user_data}"
+  user_data = "${data.ignition_config.s3.rendered}"
```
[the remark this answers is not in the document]
corest: Yes, with 100 bytes :D

modules/aws/vault/variables.tf
```diff
@@ -53,6 +57,11 @@ variable "volume_size_etcd" {
   default = 10
 }
 
+variable "volume_size_logs" {
+  type    = "string"
+  default = 5
```
kopiczko: Is it in GB?

What will happen when this volume becomes full?

corest: Yes, it is in gigabytes. If it is full - result will be the same as previously - vault will fail.

kopiczko: So why separate partition? So we can SSH?

Do you want to make it bigger afterwards?

corest: Yes, we can adjust it later. But keeping logs separately is good idea in general.

PR #235, opened 2019-04-12
Add configurable mtu for Calico

ferrandinand:
In order to use Jumbo frames we want to make MTU configurable as a parameter that can be modified by each installation.

Towards: https://github.com/giantswarm/giantswarm/issues/2117

https://docs.projectcalico.org/v3.5/usage/configuration/mtu

### DIFF
--- a/platforms/azure/giantnetes/variables.tf
+++ b/platforms/azure/giantnetes/variables.tf
@@ -181,6 +181,11 @@ variable "master_id" {
 
 ### Network ###
 
+variable "calico_mtu" {
+  description = "MTU used for Calico interfaces"
+  default     = "1500"
+}
+
 variable "vnet_cidr" {
   description = "CIDR for VMs internal virtual network."
   default     = "10.0.0.0/16"
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -34,6 +34,8 @@ storage:
             # essential.
             typha_service_name: "none"
 
+            veth_mtu: "{{.CalicoMTU}}"
+
             # The CNI network configuration to install on each node.  The special
             # values in this config will be automatically populated.
             cni_network_config: |-
@@ -45,8 +47,8 @@ storage:
                     "type": "calico",
                     "log_level": "info",
                     "datastore_type": "kubernetes",
-                    "nodename": "__KUBERNETES_NODE_NAME__",
-                    "mtu": 1500,
+                    "nodename": "__KUBERNETES_NODE_NAME__",          
+                    "mtu": __CNI_MTU__,
                     "ipam": {
                       "type": "host-local",
                       "subnet": "usePodCidr"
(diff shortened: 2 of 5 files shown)


platforms/azure/giantnetes/variables.tf
```diff
@@ -181,6 +181,11 @@ variable "master_id" {
 
 ### Network ###
 
+variable "calico_mtu" {
+  description = "MTU used for Calico interfaces"
+  default     = "1500"
```
kopiczko: Why wouldn't we switch to jumbo frames in all installations?

ferrandinand: Not all providers allow Jumbo frames, only AWS so far.

It could be also a hypothetical case where a customer workload wouldn´t be a perfect match (thinking on latency when you need to complete a full 9000 package). Not sure if it is a real scenario but I guess it is worth to be prepared.

templates/master.yaml.tmpl
```diff
@@ -45,8 +47,8 @@ storage:
                     "type": "calico",
                     "log_level": "info",
                     "datastore_type": "kubernetes",
-                    "nodename": "__KUBERNETES_NODE_NAME__",
-                    "mtu": 1500,
+                    "nodename": "__KUBERNETES_NODE_NAME__",          
+                    "mtu": __CNI_MTU__,
```
corest: where is this comes from?

ferrandinand: I saw this in calico docs:

> Note: If using Kubernetes self-hosted manifests, you should modify the veth_mtu value in the Calico ConfigMap instead, and leave "mtu" here set to __CNI_MTU__.

And it is configured here for azure (for aws was already defined)
https://github.com/giantswarm/giantnetes-terraform/blob/enable_configurable_mtu/templates/master.yaml.tmpl#L37

platforms/azure/giantnetes/variables.tf
kopiczko: So how about setting jubmo for all AWS installations?

I personally believe that we should only create knobs when we need them and not other way around.

templates/master.yaml.tmpl
corest: ack, thx

platforms/azure/giantnetes/variables.tf
ferrandinand: I guess maybe we will probably change it in all aws installations but I thought It is better to extract this value to a variable and make it configurable in just one place for all environments instead of requiring going into the template.Then if it is the case, change the default value.

I had in mind also that when enabled CP with Jumbo frames will cohabit with not upgraded tenant clusters using common MTU. This should not be a problem but not 100% sure.

@giantswarm/sre

kopiczko: Actually you are right. It's better than changing it it multiple places.

kopiczko: Sorry. This is in single place :P 

It isn't a blocker for me. I think if this is a value used in a single place it's better to keep it like it was. I'll leave the decision to the guys working more with terraform.

PR #236, opened 2019-04-16
Add crosssubnets feature

ferrandinand:
Towards: https://github.com/giantswarm/giantswarm/issues/2117


calvix:
is this still valid?

corest:
@ferrandinand ^

ferrandinand:
No, we did this to test performance for STK. This is only useful for multi A-Z deployments and would require also disable src/dest check.

PR #239, opened 2019-05-13
Disable kms auto-unseal aws

mazzy89:
During the provision of `giraffe` kms endpoint and other stuff were created but in AWS China is not possible so we have to flag the creation

Tested in giraffe

### DIFF
--- a/modules/aws/vault/vault-iam.tf
+++ b/modules/aws/vault/vault-iam.tf
@@ -1,7 +1,6 @@
 resource "aws_iam_role" "vault" {
-  count = "${var.vault_auto_unseal ? 1 : 0}"
-  name  = "${var.cluster_name}-vault"
-  path  = "/"
+  name = "${var.cluster_name}-vault"
+  path = "/"
 
   lifecycle {
     create_before_destroy = true
@@ -48,9 +47,8 @@ resource "aws_iam_role_policy" "vault-kms-unseal" {
 }
 
 resource "aws_iam_role_policy" "vault-s3-ignition" {
-  count = "${var.vault_auto_unseal ? 1 : 0}"
-  name  = "${var.cluster_name}-vault-s3-ignition"
-  role  = "${aws_iam_role.vault.id}"
+  name = "${var.cluster_name}-vault-s3-ignition"
+  role = "${aws_iam_role.vault.id}"
 
   policy = <<EOF
 {
--- a/modules/aws/vault/vault-iam.tf
+++ b/modules/aws/vault/vault-iam.tf
@@ -1,6 +1,7 @@
 resource "aws_iam_role" "vault" {
-  name = "${var.cluster_name}-vault"
-  path = "/"
+  count = "${var.vault_auto_unseal ? 1 : 0}"
+  name  = "${var.cluster_name}-vault"
+  path  = "/"
 
   lifecycle {
     create_before_destroy = true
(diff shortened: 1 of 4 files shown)


modules/aws/vault/vault-iam.tf
```diff
@@ -1,6 +1,7 @@
 resource "aws_iam_role" "vault" {
-  name = "${var.cluster_name}-vault"
-  path = "/"
+  count = "${var.vault_auto_unseal ? 1 : 0}"
```
corest: we do have `IAM` in aws china, why do you want to disable it?

modules/aws/vault/vault-iam.tf
```diff
@@ -38,14 +41,16 @@ data "aws_iam_policy_document" "vault-kms-unseal" {
...
+  count  = "${var.vault_auto_unseal ? 1 : 0}"
   name   = "${var.cluster_name}-vault-kms-unseal"
   role   = "${aws_iam_role.vault.id}"
   policy = "${data.aws_iam_policy_document.vault-kms-unseal.json}"
 }
 
 resource "aws_iam_role_policy" "vault-s3-ignition" {
-  name = "${var.cluster_name}-vault-s3-ignition"
-  role = "${aws_iam_role.vault.id}"
+  count = "${var.vault_auto_unseal ? 1 : 0}"
```
corest: same for s3. I think it is completely fine to provision vault in the same way as in other regions

modules/aws/vault/vault-iam.tf
```diff
@@ -66,8 +71,9 @@ EOF
 }
 
 resource "aws_iam_instance_profile" "vault" {
-  name = "${var.cluster_name}-vault"
-  role = "${aws_iam_role.vault.name}"
+  count = "${var.vault_auto_unseal ? 1 : 0}"
```
corest: same as above

modules/aws/vault/vault.tf
```diff
@@ -26,7 +26,7 @@ resource "aws_instance" "vault" {
   instance_type = "${var.instance_type}"
 
   associate_public_ip_address = false
-  iam_instance_profile        = "${aws_iam_instance_profile.vault.id}"
+  iam_instance_profile        = "${var.vault_auto_unseal ? aws_iam_instance_profile.vault.id : ""}"
```
corest: same as above

modules/aws/vault/vault-iam.tf
mazzy89: Here the reason why I disable this for China. 

I've seen in the commit that these TF resources have been during the KMS Auto Unseal feature. 

Hence I thought that in case auto unseal is disabled like it must be in China these resources are not necessary and hence can be disabled.

If it is not the case then please can you provide here which is the case to have IAM Role when Auto Unseal is disabled?

corest: Yes, those resources are not required. But all of them were added non because of autounseal, but because cloudconfig size was increased (auto-unesal vault config part was added). And that required using same approach for user-data of the vm as with other nodes. So I think it is good to keep this aligned even if user-data is not going beyond size limit. Otherwise we will need to bring those resources back in case we added any other small piece of configuration for vault nodelater

mazzy89: Now it makes more sense. 👍 for the insights. Then I made again the change

mazzy89:
Can this be merged?

PR #242, opened 2019-06-06
upgrade-to-1-14

calvix:
upgrade k8s to 1.14
upgrade calico to 3.6.1
upgrade etcd to 3.13
fix labels for k8s 1.14 https://github.com/giantswarm/giantswarm/issues/5560

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -1421,7 +1421,7 @@ storage:
                 serviceAccountName: kube-proxy
                 containers:
                 - name: kube-proxy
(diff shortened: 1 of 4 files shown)


calvix:
the last issue was one flag that was removed from API< this was successfully tested on gauss, we should be good here

templates/master.yaml.tmpl
```diff
@@ -2495,7 +2495,6 @@ storage:
                 - --runtime-config=api/all=true
                 - --logtostderr=true
                 - --profiling=false
-                - --repair-malformed-updates=false
```
calvix: this flag does not exist anymore

PR #246, opened 2019-06-20
user owner variable

calvix:
to fix problems in AWS china


calvix:
azure CI is fucky as always, this change is for AWS so proceeding as AWS is green

PR #247, opened 2019-06-21
Upgrade azurerm provider to latest version and fix azure e2e

calvix:
fixing azure vault , this failed e2e everytime

i accidentally sneaked changed i wanted to do in separate PR with CoreOS owner ID, but don't want to rebuild whole PR again because of the slow e2e tests, sorry :)


calvix:
feels_good

PR #250, opened 2019-07-05
Fix labeling loop

corest:
Fix infinite loop


templates/master.yaml.tmpl
```diff
@@ -2176,11 +2176,16 @@ storage:
             for namespace in ${NAMESPACES}
             do
                 while
```
kopiczko: `if` again.

PR #252, opened 2019-07-10
Add further deletion info

glitchcrab:
This ensures no dangling resources remain.


glitchcrab:
I haven't run the e2e test as it's only a minor change.

corest:
I’d request review from sig infra. Sig ops is more processes oriented

PR #257, opened 2019-08-28
Add k8s subdomain for external k8s api access

corest:
Towards https://github.com/giantswarm/giantswarm/issues/6467

To make it possible providing customers read-only access to control-plane API, we need expose somehow API access on network level. Currently it is behind private balancer, which is available only via VPN. 

This can be solved by adding ingress resource with passthrough. Used domain in ingress must be allowed in alt_names of the api certificates. So this PR allows it.

Not sure about naming, maybe it should be something explicit like `external-k8s-api.g8s.<base-domain>`. Thoughts?

### DIFF
--- a/platforms/aws/giantnetes/main.tf
+++ b/platforms/aws/giantnetes/main.tf
@@ -96,6 +96,7 @@ locals {
     "ExternalVpnVultrPassword"     = "${var.external_ipsec_password}"
     "ExternalVpnVultrSubnet"       = "${var.external_ipsec_subnet_1}"
     "ExternalVpnVultrSourceIp"     = "${cidrhost("${var.external_ipsec_subnet_1}",1)}"
+    "ExternalAPIDomainName"        = "${var.external_api_dns}.${var.api_dns}.${var.base_domain}"
     "G8SVaultToken"                = "${var.nodes_vault_token}"
     "ImagePullProgressDeadline"    = "${var.image_pull_progress_deadline}"
     "K8SAPIIP"                     = "${var.k8s_api_ip}"
--- a/platforms/aws/giantnetes/variables.tf
+++ b/platforms/aws/giantnetes/variables.tf
@@ -133,6 +133,12 @@ variable "api_dns" {
   default     = "g8s"
 }
 
+variable "api_external_dns" {
+  type        = "string"
+  description = "External FQDN for api (i.e. k8s)."
+  default     = "k8s"
+}
+
 variable "etcd_dns" {
   type        = "string"
   description = "FQDN for etcd (i.e. etcd). '__MASTER_ID__' wil be replaced with id of master"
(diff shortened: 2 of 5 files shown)


corest:
Example of the ingress, which will be used for external access

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: k8s
  namespace: default
  labels:
    kubernetes.io/ingress.class: nginx.ingress.kubernetes.io
  annotations:
    nginx.ingress.kubernetes.io/ssl-passthrough: "true"
    kubernetes.io/ingress.class: "nginx"
    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
    kubernetes.io/tls-acme: "true"
    certmanager.k8s.io/cluster-issuer: "letsencrypt"
spec:
  tls:
  - hosts:
    - k8s.g8s.gauss.eu-central-1.aws.gigantic.io
    secretName: k8s-certs
  rules:
  - host: k8s.g8s.gauss.eu-central-1.aws.gigantic.io
    http:
      paths:
      - backend:
          serviceName: kubernetes
          servicePort: 443
        path: /
```

corest:
What about `external-k8s`?

calvix:
Why we would need to add the alternative name? This worked nicely without it.

The Nginx is connection to the IP and that ips is already in IP sans,  and certs for  the nginx ingress are generated by LE

corest:
I know it works fine with tls over Ingress without passthrough. As you’ve mentioned - this thing is done in this way because of cert authorization. I’m just following guidelines - sa is for application, not for humans. So if GS staff is using certs, external people also using certs(created with single opsctl command in similar way as for us) and we’re not mixing humans and apps. That’s debatable, sure. More thoughts needed here

calvix:
Not really sure if service-account is only for apps, but anyway I have no strong opinion here, just thought that it would be easier to manage them.

corest:
Read that here https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#user-accounts-vs-service-accounts

```
User accounts are for humans. Service accounts are for processes, which run in pods.
```

Maybe @puja108 also have something to add as he originally proposed using SAs

puja108:
I've seen people use SA for users as it is has a bit of a security advantage (revocation as well as SA CA rotation). The group problem can be solved with a trick (not super elegant) where you create a reserved namespace and bind the role to all SAs in that namespace. Now when you create an SA in that namespace it automatically has that role. SA generation can also very easily be automated as it is just 2 API calls (1 for creation and 1 for getting the token).

puja108:
If we anyway want to go with an additional domain I would like to chose something that stays with us and makes it more clear. Maybe something like `gs-clusterapi` to show somehow the relation to upstream?

puja108:
One thing I fear with certificate management is that we will at some point revamp the whole thing. Another thing that we should keep in mind is that at some point customers will have a requirement for secure access to the cluster api, which for our customers means SSO with MFA. While that is for the future we should keep that in the back of our heads when thinking about this.

corest:
Thanks @calvix and @puja108. Your points are valid for me. I'll proceed with SA tokens

corest:
p.s. `gs-clusterapi` as dns for ingress is still a valid thing. thx @puja108

PR #259, opened 2019-08-29
Add ingress/rbac for external access

corest:
Towards https://github.com/giantswarm/giantswarm/issues/6813

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -1741,6 +1741,7 @@ storage:
             - nodes
             verbs:
             - list
+          ---
           ## CoreDNS
           apiVersion: rbac.authorization.k8s.io/v1
           kind: ClusterRole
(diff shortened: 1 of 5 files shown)


templates/master.yaml.tmpl
```diff
@@ -1695,6 +1725,24 @@ storage:
...
+            namespace: giantswarm-clusterapi
+            labels:
+              scope: control-plane
+              type: external
+              level: readonly
+          rules:
+          - apiGroups:
+            - ""
+            resources:
+            - nodes
```
calvix: I guess this is start and will be adjusted later

corest: Yep, this is just a base for tests

templates/master.yaml.tmpl
```diff
@@ -1676,6 +1706,7 @@ storage:
             labels:
               k8s-app: nginx-ingress-controller
           spec:
+            externalTrafficPolicy: Local
```
corest: This one is tricky a bit. For services of type `NodePort` or `LoadBalancer` filtering works only with `Local` external traffic policy. In this mode source IP is not changed. That in general affects traffic balancing as requires  nginx ic on every worker. But in case of control-plane that is the case - every worker has IC pod. So no harmful impact here. Onprem we don't need all of this as customers anyway have network access to control-plane API.

Just bringing this for awareness of @giantswarm/sig-infrastructure

corest:
Waiting for CI and merging

PR #263, opened 2019-09-05
Use native templating -  poc for bastions

corest:
Towards https://github.com/giantswarm/giantswarm/issues/6858

Template is splitted into files

gotemplate plugin is extended with nested templates support


platforms/aws/giantnetes/main.tf
```diff
@@ -119,15 +119,35 @@ locals {
   }
 }
 
-# Generate ignition config.
-data "gotemplate" "bastion" {
-  template = "${path.module}/../../../templates/bastion.yaml.tmpl"
-  data     = "${jsonencode(local.ignition_data)}"
+# Render files before referencing them in main state
+# It should be applied before main apply
+# terraform apply -target="module.ignition" ./
```
corest: this will be added into bootstrap.sh in installations repository. I didn't find any other way how to overcome this.

calvix:
another idea would be use of some terraform plugin which would  do the magic behind templating and we could still have template files

corest:
Would be cool to have templates_dir in go-template plugin. That's one of the main reasons I'm switching this. Writing `gotemplate` plugin calls for every file template isn't a super good thing. So I could implement that inside plugin. Native `templates_dir` generates the same files structure for rendered files automatically.

But then all the templating stuff is usually specific to the platform. So we're not writing templates that often. It is usually about editing existing files. And that is easier to do with raw files. 

Anyway, waiting for one more lgtm and will proceed with another templates tomorrow

platforms/aws/giantnetes/main.tf
```diff
@@ -119,15 +119,35 @@ locals {
...
+module "ignition" {
+  source = "../../../modules/ignition"
+
+  ignition_data = "${local.ignition_data}"
+}
+
+# List with base64 encoded files, injected into main templates
+locals {
+  Files = {
+      "conf/hardening.conf" = filebase64("../../../files/conf/hardening.conf")
```
ferrandinand: Not stopper, but can we use ${path.module} instead of ../../../ ?
Seems cleaner and we protect us against any modification in the code structure

ferrandinand:
Hard to say: 
I don´t want to have 2 different templating sintax so we are still diverging ansible installation vs terraform installation but seems that terraform code can cleaner.
I don´t like neither the need to run a terraform apply before but once diverged in a future if we see it is painful we can start to use ignition provider instead of ct_config. 

Once said this, I prefer native templating if we decide that having 2 different systems is not a stopper.

corest:
There are few important points: 
  - writing terraform template is less annoying than editing yaml files inside string templates
  - same files can be reused across all templates (e.g. I don't need edit sshd_config with deprecated parameter in 8 files inside string templates)
  - we're templating stuff rarely

corest:
Dropped single master tests.

corest:
AWS:
```
stage-wait-kubernetes-nodes: Expected nodes 4, actual nodes 4.
stage-wait-kubernetes-nodes: Kubernetes nodes are ready.
```

Azure
```
stage-wait-kubernetes-nodes: Expected nodes 4, actual nodes 4.
stage-wait-kubernetes-nodes: Kubernetes nodes are ready.
```

PR #264, opened 2019-09-16
PSP for calico-kube-controllers

glitchcrab:
towards giantswarm/giantswarm/issues/6013

This PR still allows the pod to run as root as this is required to read certs from the hostPath dir. A later PR should address this.

This has been tested by manually applying the role/bindings/psp on gauss.

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -1093,6 +1093,40 @@ storage:
 
           ---
 
+          # This manifest creates a clusterrole for the PSP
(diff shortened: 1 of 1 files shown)


corest:
What is this for? We're already binding calico to privileged psp https://github.com/giantswarm/giantnetes-terraform/blob/master/templates/master.yaml.tmpl#L2244-L2250

corest:
We probably can have separate. But while all of those request wide range of stuff, we can keep current binding

PR #266, opened 2019-10-03
add k8s-api-healthz service

calvix:
towards https://github.com/giantswarm/giantswarm/issues/6911

adding k8s api healthz micro service for LB health checks

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -2726,7 +2726,15 @@ storage:
             priorityClassName: system-node-critical
             containers:
               - name: k8s-api-healthz
-                image: {{.DockerRegistry}}/giantswarm/k8s-api-healthz:6d8a2834361d336524702b0b2e485a75cc07652d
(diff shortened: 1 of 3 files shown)


calvix:
All green, the Azure CI only fails in the end in the deletion process. but all nodes are up and healthy

templates/master.yaml.tmpl
```diff
@@ -2709,6 +2709,45 @@ storage:
...
+        inline: |
+          apiVersion: v1
+          kind: Pod
+          metadata:
+            name: k8s-api-healthz
+            namespace: kube-system
+            annotations:
+              scheduler.alpha.kubernetes.io/critical-pod: ''
+          spec:
+            hostNetwork: true
```
kopiczko: Is it because it's potentially deployed before calico?

calvix: Yes, its static pod it needs to run without k8s-api  at all

PR #270, opened 2019-10-16
Bump azure k8s version to 1.15.5

MarcelMue:
From changelog:
```
Fix aggressive VM calls for Azure VMSS
```
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#changelog-since-v1154


ferrandinand:
Maybe better to run the pipeline with the branch

PR #272, opened 2019-10-18
Mount lib-modules in kube-proxy

ferrandinand:
Towards: https://github.com/giantswarm/giantswarm/issues/7273

Already present in hive and k8scloudconfig

### DIFF
--- a/templates/files/k8s-resource/kube-proxy-ds.yaml
+++ b/templates/files/k8s-resource/kube-proxy-ds.yaml
@@ -67,6 +67,9 @@ spec:
         - mountPath: /etc/kubernetes/ssl
           name: ssl-certs-kubernetes
           readOnly: true
+        - mountPath: /lib/modules
+          name: lib-modules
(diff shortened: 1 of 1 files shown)


JosephSalisbury:
What is `/lib/modules`?

corest:
Maybe it makes sense to check why it takes that much time? It worked with previous tests and adding volume should not cause longer startup

ferrandinand:
Lib modules is all compiled kernel modules for the underlying kernel in the OS.
This error is known here https://github.com/kubernetes/kubernetes/issues/55043

Increasing the the number of retries is more about checking that this is not breaking anything and it is only a matter of time.Increasing the retries make it to success.

I see errors in waiting for nodes up from long time ago in Azure and AWS so I would create a separate issue in SIG-infra to discuss about this a check why is it getting too much time and how to improve it.

ferrandinand:
I don´t know if it is taking so much

PR #275, opened 2019-10-23
Upgraded calico to version 3.9.2 for azure

whites11:


calvix:
cool, merge it

templates/files/manifests/controller-manager.yaml
```diff
@@ -68,4 +71,7 @@ spec:
     name: k8s-kubeconfig
   - hostPath:
       path: /etc/kubernetes/secrets
-    name: k8s-secrets
\ No newline at end of file
+    name: k8s-secrets
+  - hostPath:
+      path: /var/libexec/kubernetes/kubelet-plugins/volume/exec
```
calvix: I would still rather have this under `/var/lib/kubelet/plugins/volume/exec` as that is the place where rest of kubelet files are and its also separate volume than root volume

whites11:
e2e tests are now consinstently passing.
Tested on godsmack and ghost and everything looks good.
Merging.

PR #279, opened 2019-11-19
Upgrade to calico 3.9.3 for AWS

whites11:
See https://github.com/giantswarm/giantswarm/issues/7519

### DIFF
--- a/templates/files/k8s-resource/calico-all.yaml
+++ b/templates/files/k8s-resource/calico-all.yaml
@@ -751,17 +751,36 @@ subjects:
 {{- end }}
 
 {{ if eq .Provider "aws" -}}
-# Extra changes:
-#  - Added resource limits to calico-node and calico-kube-controllers.
-#  - Added resource limits to install-cni.
-#  - Added critical pod priority to calico-node daemonSet and calico-kube-controllers 'priorityClassName: system-cluster-critical'.
-#
-# Calico Version v3.6.1
-# https://docs.projectcalico.org/v3.6/releases#v3.6.1
+# Source # https://docs.projectcalico.org/v3.9/manifests/calico-etcd.yaml
+# Changed values:
+#   - Adjusted all image references to giantswarm docker registry
+#   - In calico-config ConfigMap:
+#     - Set etcd_endpoints to "{{.ETCDEndpoints}}"
+#     - Set etcd_ca to "/calico-secrets/etcd-ca"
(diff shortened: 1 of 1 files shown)


templates/files/k8s-resource/calico-all.yaml
```diff
@@ -1046,6 +1069,13 @@ spec:
               name: cni-net-dir
             - mountPath: /calico-secrets
               name: etcd-certs
+        # Adds a Flex Volume Driver that creates a per-pod Unix Domain Socket to allow Dikastes
+        # to communicate with Felix over the Policy Sync API.
+        - name: flexvol-driver
+          image: {{.DockerRegistry}}/giantswarm/pod2daemon-flexvol:v3.9.3
+          volumeMounts:
+            - name: flexvol-driver-host
+              mountPath: /host/driver
```
ferrandinand: What is the intention behind of using pod2daemon-flexvol? 
In the comment say to allow Dikastes communicate with Felix but I don´t see dikastes images.

If it is all intended, it could be great to add some info or link in the PR description about new things to understand what are we changing

whites11: Honestly I have no idea.
The pod and the comment come from the official .yaml: https://docs.projectcalico.org/v3.9/manifests/calico-etcd.yaml

tuommaki: I think this should be checked and documented/communicated. It's not fully clear what this or Dikastes does, but what I can find from scarce resources in Calico documentation and Google searches, it seems to be a side-car container used with Envoy in Istio setups and potentially related to network policy configurations. I'm not sure if we should have something like this enabled out of the box, but I'm also wondering if this is requirement when customer wants to run Istio on top of our clusters.

Ping @pipo02mix - do you potentially know more about this wrt. Istio?

pipo02mix: Here is well explained https://github.com/projectcalico/app-policy

So it is not only for IStio (you can use to add policies at layer 7 for your apps) but Istio makes it easier to use. I would like to have it, but I would do it in a different story (PR)

corest:
@mazzy89 can you provide info about the current state in tenant clusters?

mazzy89:
Ludacris team is working on the next release and wiring a PR to upgrade Calico here https://github.com/giantswarm/k8scloudconfig/pull/611

PR #281, opened 2019-11-21
Use explicit file name for master/worker k8s cloudprovider config for Azure

fiunchinho:
I renamed the files to be more explicit about which one is used for `master`s and which one for `worker`s. Also, both moved to the same folder.


corest:
Merge? @fiunchinho

fiunchinho:
Sorry, I forgot about this. I'm not sure this change will affect anything else. Do I have to update any dependency? Otherwise, let's merge it.

corest:
Just merge

PR #285, opened 2019-12-11
Always deploy the delegated DNS zone in terraform

whites11:
Fixed a bug in the terraform module "dns".
See comments in the code.

### DIFF
--- a/modules/azure/dns/dns.tf
+++ b/modules/azure/dns/dns.tf
@@ -23,7 +23,7 @@ variable "zone_name" {
 }
 
 resource "azurerm_dns_zone" "dns_zone" {
-  count               = "${var.root_dns_zone_name == "" ? 0 : 1}"
+  count               = "1"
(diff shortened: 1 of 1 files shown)


modules/azure/dns/dns.tf
```diff
@@ -23,7 +23,7 @@ variable "zone_name" {
 }
 
 resource "azurerm_dns_zone" "dns_zone" {
-  count               = "${var.root_dns_zone_name == "" ? 0 : 1}"
+  count               = "1"
```
whites11: This resource is the one that creates a DNS zone in the installation's resource group and needs to be created always.
Using the `root_dns_zone_name` variable to enable or disable the creation of this delegated zone is wrong and leads to an error during the terraform execution.
The variable should be used to just control the creation of the NS records in the delegating zone (such as `azure.gigantic.io`).

PR #286, opened 2020-01-02
Add cluster-type tag

corest:
This PR https://github.com/giantswarm/aws-operator/pull/2062/files introduced lookup for a control-plane VPC, which works only when there are separate accounts used for control-plane and tenant in aws. In China we have a single account, so filter from here https://github.com/giantswarm/aws-operator/blob/master/service/controller/resource/cpvpc/resource.go#L100-L127 returns multiple records. 

With additional tag it will be easier to get control-plane vpc id


corest:
After merge I'll manually tag all the control-plane vpcs and then add a filter into aws-operator

PR #287, opened 2020-01-06
Use proper docker registry for CP

calvix:
towards https://github.com/giantswarm/giantswarm/issues/7435

certctl was still using the quay image even in china

### DIFF
--- a/templates/files/conf/k8s-addons
+++ b/templates/files/conf/k8s-addons
@@ -1,7 +1,8 @@
 #!/bin/bash
 set -eu
-# k8s 1.14.1
-KUBECTL_IMAGE=quay.io/giantswarm/docker-kubectl:1.14.1
+# k8s 1.16.4
+# WHEN UPDATING TAG DONT FORGET TO UPDATE VERSION ABOVE AS WELL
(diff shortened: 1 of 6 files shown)


calvix:
> LGTMIIWFY; Is `943e40d9c36efc2eec76783d48a891fc6f323493` latest `certctl`?

Yep, but I decided to extend this PR to include all components. Sorry for that

templates/files/conf/k8s-addons
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 set -eu
-# k8s 1.14.1
-KUBECTL_IMAGE=quay.io/giantswarm/docker-kubectl:1.14.1
+# k8s 1.16.4
+# WHEN UPDATING TAG DONT FORGET TO UPDATE VERSION ABOVE AS WELL
+KUBECTL_IMAGE={{.DockerRegistry}}/giantswarm/docker-kubectl:4377f52eeb22d766bbc59bdf71ce0724af61a4f6
```
calvix: decided to go back to tags, as we don't have version tag created by default and this would mean someone has to do taht manually in china and that would probably get forgotten

calvix:
fking non root user

PR #288, opened 2020-01-21
Adjust TF with 0.12.19 version

corest:
Towards https://github.com/giantswarm/giantswarm/issues/8507


corest:
CI for azure randomly fails on accessing API. Tested manually, no issues

PR #289, opened 2020-02-11
Refactor network with public/private elbs

corest:
Towards https://github.com/giantswarm/giantswarm/issues/8842

This allows us to get rid of port forwarding

Align with tenant cluster architecture


corest:
Deployed to gauss. Works fine

PR #290, opened 2020-02-12
Use asg for masters

calvix:
use ASG for masters instead of single instance
This same setup will be used for the HA masters for TC

changes:
* added ENI that is mouted to the master instance for static ip for the ETCD , ENI always takes first usable IP from the subnet (0 - network, 1 gateway, 2 -3 reserved by AWS, 4 - eni etcd)
* add etcd3-attach-dep service which is https://github.com/giantswarm/aws-attach-etcd-dep/pull/4 this will attach etcd EBS volume and the ENI during bootup
* move etcd listen ip to the ENI ip
* etcd is now mounted via `automount` that measn it mounted when someone tries to use that path
*


modules/aws/master/master.tf
```diff
@@ -7,62 +7,95 @@ locals {
...
 data "aws_availability_zones" "available" {}
 
-resource "aws_instance" "master" {
+
+resource "aws_cloudformation_stack" "master_asg" {
+  count = var.master_count
+  name  = "${var.cluster_name}-master-${count.index}"
+
+  template_body = <<EOF
+{
```
mazzy89: why having CF here?

calvix: We had the same CF for workers, I think the reason was that CF is handling the ASG update somehow better than terraform but maybe it was due to some old bug

calvix: I will try use ASG directly

calvix: so I tried to use ASG directly but somehow the change of ignition does not trigger ASG rolling update, no idea why.

So that is probabbly reason why we use CF which does the magic of rolling machines for us

PR #294, opened 2020-02-19
K8s 1.16

calvix:
update many components but mainly k8s version
used k8scloudconfig as a template for changes from containerized kubelet


calvix:
I don't think there are changes necessary, ports fro monitoring stay the same.

calvix:
checked prom on gauss and it looks happy

PR #301, opened 2020-02-28
Fix deployment specs for new k8s

corest:
This was missed during the update to k8s 1.16 as those resources were already there.

P.S. @calvix maybe worth to check hive repo as well as you would get into same issue there with new installation


calvix:
@corest Yep I have the have repo fixed now in the PR, somehow I was looking for k8s-addons there but forgot to check it for terraform :(

PR #305, opened 2020-03-13
update k8s to 1.17

calvix:
Towards https://github.com/giantswarm/giantswarm/issues/11399

update k8s to 1.17 , lets back on on the track !


corest:
Updated this PR,
Deployed to ginger and godsmack (aws/azure)
Works fine
will request a review in the morning

corest:
Both CI fail due to issue with waiting for calico. Will investigate now

calvix:
Cant approve, but LGTM if it works

modules/azure/vault/vault.tf
```diff
@@ -40,6 +40,12 @@ resource "azurerm_virtual_machine" "vault" {
     version   = var.flatcar_linux_version
   }
 
+  plan {
+    name = var.flatcar_linux_channel
+    publisher = "kinvolk"
+    product = "flatcar-container-linux-free"
+  }
```
kopiczko, line 47: Why do we need this `plan` block now?

corest: It is related to Kinvolk providing licensed versions of Flatcar via marketplace. @whites11 can give more details

whites11: In theory we should not need this (this is normally used on Azure to get access to paid images, but this one is free).
We got in touch with Kinvolk to get that fixed, but they didn't solve it yet.
Therefore we need to keep pushing that block in all our API requests that use the flatcar linux free image.

PR #308, opened 2020-03-17
migrate-to-aws-cni

calvix:
towards : https://github.com/giantswarm/giantswarm/issues/9259
migrate from calico CNI to AWS CNI

TODO:
* make one calico-policy manifest for both Azure and AWS


calvix:
the new cluster works nicely, I need to test it on our running installation now

calvix:
tested on gauss. migration succesfull

calvix:
Will add some info to the installation guide

modules/aws/vpc/vpc-subnet-cni.tf
```diff
@@ -0,0 +1,56 @@
+resource "aws_vpc_ipv4_cidr_block_association" "cni" {
+  vpc_id     = aws_vpc.cluster_vpc.id
+  cidr_block = var.aws_cni_cidr_block
+}
+
```
calvix: using VPC cird clock association to be as close to TC as possible ( also  increasing VPC size is not really possible so this could be a problem for already existing installations)

PR #310, opened 2020-04-05
Switched to externalTrafficPolicy Cluster 

whites11:
The starting point of this funny bug is this PM: https://github.com/giantswarm/giantswarm/issues/9775

The initial symptom: nginx ingress controller not reachable in the control plane when the pods were all scheduled in the master nodes and none in the worker nodes.

The problem is that we use `externalTrafficPolicy: Local` which causes `kube-proxy` to avoid doing NAT on incoming requests. Those requests will fail if the POD is not running in the node where the request is arrived and this ultimately makes the azure load balancer think the node is unhealthy.

Not sure if AWS control planes work and how they do that.

k8s sig network people tells this is intended behaviour, so we have no choice but switch to `externalTrafficPolicy: Cluster` https://github.com/kubernetes/kubernetes/issues/89831

### DIFF
--- a/templates/files/conf/k8s-addons
+++ b/templates/files/conf/k8s-addons
@@ -125,7 +125,7 @@ MANIFESTS="default-storage-class.yaml
  default-backend-svc.yaml
  ingress-controller-cm.yaml
  ingress-controller-sa.yaml
- ingress-controller-dep.yaml
+ ingress-controller-daemonset.yaml
  ingress-controller-svc.yaml"
 for manifest in $MANIFESTS
 do
--- a/templates/files/k8s-resource/ingress-controller-daemonset.yaml
+++ b/templates/files/k8s-resource/ingress-controller-daemonset.yaml
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: Deployment
+kind: DaemonSet
 metadata:
   name: nginx-ingress-controller
   namespace: kube-system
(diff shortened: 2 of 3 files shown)


corest:
Im not completely sure about this solution. Why don’t we just run ic on workers(those nodes which are ingress balancer pool)?
With the cluster external traffic policy we will lose source ip of the client. How then ip whitelisting gonna work?

whites11:
> Im not completely sure about this solution. Why don’t we just run ic on workers(those nodes which are ingress balancer pool)?
> With the cluster external traffic policy we will lose source ip of the client. How then ip whitelisting gonna work?

Yes that's a good idea if we care about the client IP.
I am not even sure why the pods are scheduled on the masters at all.
Will investigate.
THX for feedback

whites11:
Ok the reason is that the master nodes have no taints, thus they're considered normally schedulable.
I switched from a deployment to a daemonset, with a nodeSelector to schedule only on worker nodes.

WDYT?

corest:
Switch to daemonsets will work. 
But do we need that many pods of nginx ic?
Also how the whole thing worked before? We need to find what has changed

whites11:
> Switch to daemonsets will work.
> But do we need that many pods of nginx ic?

Considering the NodePort service we need to have one running for each worker node, or Azure will go nuts. So yes, we need that many pods for now.

> Also how the whole thing worked before? We need to find what has changed

Right. If that's ok I will merge this PR to mitigate the issue but leave the PM open for further investigation.

corest:
> Considering the NodePort service we need to have one running for each worker node, or Azure will go nuts. So yes, we need that many pods for now.

We have 3 replicas for 3 workes and preferred scheduling for workers, so that worked without daemonset before

PR #313, opened 2020-04-13
Add K8S API LB whitelisting

corest:
Towards https://github.com/giantswarm/giantswarm/issues/10041


corest:
CI is green, tested on `gauss`/`godsmack` installations - works fine

PR #317, opened 2020-04-29
[aws] Flatcar Linux (no China)

mazzy89:
This PR introduces to the GS CP Flatcar.

I've opted for the current approach. If it is not set it up explicitly `flatcar_linux_version`, CoreOS is still used as OS for all CP compoments.

This covers only AWS. For Azure we need to refactor even more, because unfortunately the abstraction is not so powerful. Azure modules indeed explicitly define as publisher `CoreOS`. oh well let's leave this discussion for a different PR.

I've already executed `terraform plan` and it works. Need to run `apply` to create `galileo` as first CP with Flatcar Linux.

### DIFF
--- a/modules/flatcar-linux/data/version.sh
+++ b/modules/flatcar-linux/data/version.sh
@@ -1,0 +1,22 @@
+#!/bin/sh
+
+set -e
+
+if [ ! "$(command -v jq)" ]; then
+  echo "jq is not installed."
+  exit 1
+fi
+
+eval "$(jq -r '@sh "FLATCAR_CHANNEL=\(.flatcar_channel) FLATCAR_VERSION=\(.flatcar_version) AWS_REGION=\(.aws_region)"')"
+
+if [ "$FLATCAR_VERSION" = "latest" ]; then
+  if [ "$AWS_REGION" != "cn-north-1" ] && [ "$AWS_REGION" != "cn-northwest-1" ]; then
+    version=$(curl -s https://"$FLATCAR_CHANNEL".release.flatcar-linux.net/amd64-usr/current/version.txt | sed -n 's/FLATCAR_VERSION=\(.*\)$/\1/p')
+  else
+    version=$(curl -s https://flatcar-prod-ami-import-cn-north-1.s3.cn-north-1.amazonaws.com.cn/version.txt | sed -n 's/FLATCAR_VERSION=\(.*\)$/\1/p')
+  fi
+else
+  version="$FLATCAR_VERSION"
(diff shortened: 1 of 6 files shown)


modules/flatcar-linux/main.tf
```diff
@@ -0,0 +1,3 @@
+data "external" "flatcar_version" {
+  program = ["sh", "-c", "curl https://${var.flatcar_channel}.release.flatcar-linux.net/amd64-usr/current/version.txt | sed -n 's/FLATCAR_VERSION=\\(.*\\)$/{\"flatcar_version\": \"\\1\"}/p'"]
```
MarcelMue, line 2: I don't think this works for AWS china, correct?

mazzy89: No, that's true. For China we don't push this file format. I have probably to create it.

mazzy89: Going to add also the use case for China by pushing to S3 `version.txt`

mazzy89:
Thank you fellas. I have just included the AWS China cases here too.

modules/flatcar-linux/data/version.sh
```diff
@@ -0,0 +1,22 @@
...
+
+if [ ! "$(command -v jq)" ]; then
+  echo "jq is not installed."
+  exit 1
+fi
+
+eval "$(jq -r '@sh "FLATCAR_CHANNEL=\(.flatcar_channel) FLATCAR_VERSION=\(.flatcar_version) AWS_REGION=\(.aws_region)"')"
+
+if [ "$FLATCAR_VERSION" = "latest" ]; then
+  if [ "$AWS_REGION" != "cn-north-1" ] && [ "$AWS_REGION" != "cn-northwest-1" ]; then
```
kopiczko, line 13:
```suggestion
  if [ "$AWS_REGION" != cn-* ]; then
```

mazzy89: Unfortunately regex in the shell does not work in that way but thank you anyway for the suggestion.

mazzy89: fyi https://www.networkworld.com/article/2693361/unix-tip-using-bash-s-regular-expressions.html

kopiczko: I didn't realise you're using sh here.

mazzy89:
Thank you @corest for a test here. I haven't tested China. Going to merge this and then I'll fix it in a different PR.

PR #319, opened 2020-05-04
Add internal g8s elb

corest:
Towards https://github.com/giantswarm/roadmap/issues/128

The customer wants to access control-plane k8s API via internal elb


calvix:
didn't we had an issue with small subnets on some installations?

corest:
@calvix yep, we had an issue but as I've doubled subnet size on those installations, it should be enough. But will verify in China before merging

PR #324, opened 2020-05-06
Proper namespace creation

corest:
Removed namespace creation from doc and added heree


corest:
Tested manually few times. Works fine

PR #326, opened 2020-05-15
Use 0.0.0.0/0 when it is the only network allowed

corest:
Towards https://github.com/giantswarm/giantswarm/issues/10911

### DIFF
--- a/modules/azure/vnet/nsg-worker.tf
+++ b/modules/azure/vnet/nsg-worker.tf
@@ -1,10 +1,10 @@
 locals {
   
-  customer_vpn_subnets = var.customer_vpn_subnets != "" ? split(",", var.customer_vpn_subnets) : []
+  customer_vpn_public_subnets = var.customer_vpn_public_subnets != "" ? split(",", var.customer_vpn_public_subnets) : []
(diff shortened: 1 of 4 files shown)


modules/azure/vnet/nsg-worker.tf
```diff
@@ -73,7 +73,7 @@ resource "azurerm_network_security_rule" "master_ingress_ingress" {
 resource "azurerm_network_security_rule" "worker_ingress_internal_any" {
   name                        = "${var.cluster_name}-worker-in-any"
   description                 = "${var.cluster_name} worker - Internal"
-  priority                    = 800
+  priority                    = 900
```
corest, line 76: fixing conflicting rules priorities

platforms/azure/giantnetes/main.tf
```diff
@@ -41,7 +41,7 @@ module "vnet" {
   bastion_count       = "2"
   bastion_cidr        = var.bastion_cidr
   cluster_name        = var.cluster_name
-  customer_vpn_subnets = var.customer_vpn_subnets
+  customer_vpn_public_subnets = var.customer_vpn_public_subnets
```
corest, line 44: aligning naming with aws

PR #335, opened 2020-06-11
Added plan for `vault` VM on azure (needed for flatcar)

whites11:
Vault deployment wasn't working without this

### DIFF
--- a/modules/azure/vault/vault.tf
+++ b/modules/azure/vault/vault.tf
@@ -40,6 +40,12 @@ resource "azurerm_virtual_machine" "vault" {
     version   = var.flatcar_linux_version
   }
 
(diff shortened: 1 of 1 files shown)


corest:
@whites11 what is the state of this?

whites11:
> @whites11 what is the state of this?

I need to test it to ensure it doesn't break existing installations.
Feel free to do it if you have time.

whites11:
@corest FYI I tried to "apply" from this branch to godsmack today. Terraform tries to reconcile the difference (add the "plan" to the VM) but can't because that's unsupported by azure API.
This can't be merged IMHO and we must wait for this repo to be versioned before we can do that.

PR #336, opened 2020-06-16
reduce-aws-cni-warm-ip

calvix:
this can help when the cluster is auto-scaling a lot to speed up process but the problem is that it will waste IPs on  each node so big number can lead to unused assigned IP and in small subnets lead to exhaustion of subnets


templates/files/k8s-resource/aws-cni.yaml
```diff
@@ -136,9 +136,9 @@ spec:
             ## Deviation from original manifest - 4
             ## prewarm IP to limit AWS api calls
```
kopiczko, line 137: Is this comment still valid?

calvix: yep

calvix:
ci is somehow angry

PR #338, opened 2020-06-18
templates: add k8s-setup-delete-old-node.service

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/11235

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -1019,8 +1019,8 @@ systemd:
     contents: |
       [Unit]
       Description=Deletes old kubernetes node which this instance is going to replace
-      After=k8s-setup-download-hyperkube-binaries.service{{ if eq .Provider "aws" }} set-hostname.service{{ end }}
-      Requires=k8s-setup-download-hyperkube-binaries.service{{ if eq .Provider "aws" }} set-hostname.service{{ end }}
+      After=k8s-setup-download-hyperkube-binaries.service api-certs.service {{ if eq .Provider "aws" }} set-hostname.service{{ end }}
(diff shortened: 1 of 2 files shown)


templates/master.yaml.tmpl
```diff
@@ -1014,14 +1014,33 @@ systemd:
...
       [Install]
       WantedBy=multi-user.target
+  - name: k8s-setup-delete-old-node.service
+    enabled: true
+    contents: |
+      [Unit]
+      Description=Deletes old kubernetes node which this instance is going to replace
+      After=k8s-setup-download-hyperkube-binaries.service{{ if eq .Provider "aws" }} set-hostname.service{{ end }}
+      Requires=k8s-setup-download-hyperkube-binaries.service{{ if eq .Provider "aws" }} set-hostname.service{{ end }}
+      ConditionFirstBoot=yes
```
kopiczko, line 1024: Note it's fired only on first boot.

kopiczko: I checked it isn't run on reboots and k8s-kubelet still starts.

kopiczko:
Tested on ginger.

calvix:
this doesn't work,  when  all master nodes are down or it's a fresh installation it cannot delete node since there is no API available and this causes k8s-kubelet not start since its dependency

  we need to rollback it and then try to run CI to ensure it works on fresh clusters

kopiczko:
> this doesn't work, when all master nodes are down or it's a fresh installation it cannot delete node since there is no API available and this causes k8s-kubelet not start since its dependency
> 
> we need to rollback it and then try to run CI to ensure it works on fresh clusters

https://github.com/giantswarm/giantnetes-terraform/pull/339

calvix:
thx

PR #340, opened 2020-07-08
update-k8s-network-setup-image

calvix:
towards https://github.com/giantswarm/giantswarm/issues/11585

changes in `k8s-setup-network-env` are here https://github.com/giantswarm/giantswarm/issues/11585


calvix:
tested on aws and azure,

PR #341, opened 2020-07-13
Add a common tag to CP resources for tracking purposes

pipo02mix:
Now we have installation tag but we don't have a distinct tag between CPs and tenants. For tracking purposes it is worthy


pipo02mix:
Tested on gauss 🚀

PR #342, opened 2020-07-15
Increase the default instance-type

njuettner:
Related issue: giantswarm/giantswarm#11899

### DIFF
--- a/platforms/aws/giantnetes/variables.tf
+++ b/platforms/aws/giantnetes/variables.tf
@@ -87,7 +87,7 @@ variable "master_instance" {
   type = map
 
   default = {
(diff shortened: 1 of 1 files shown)


platforms/aws/giantnetes/variables.tf
```diff
@@ -87,7 +87,7 @@ variable "master_instance" {
   type = map
 
   default = {
-    type          = "m5.large"
+    type          = "m5x.large"
```
calvix, line 90: I think the right name is `m5.xlarge`

njuettner: Sorry my bad, typo x.x

PR #343, opened 2020-07-17
Check calico with retries

corest:
Towards https://github.com/giantswarm/giantswarm/issues/11236

For now kubectl wait hangs at some point due to timeout accessing API and never retries again. That's why it fails eventually


calvix:
`engineer brain is broken` :trollface:

PR #347, opened 2020-07-21
Ext employees

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/11732

This PR uses GS users' data from the `employees` repository for the azure platform

### DIFF
--- a/platforms/azure/giantnetes/variables.tf
+++ b/platforms/azure/giantnetes/variables.tf
@@ -3,6 +3,12 @@ variable "github_token" {
   description = "Your personal GITHUB token, used to get access to the private repository 'employees' to get the list of users."
 }
 
+variable "employees_branch" {
+  type        = string
+  description = "The branch in the 'employees' repo to use for getting a list of users"
+  default     = "master"
+}
+
 variable "cluster_name" {
   type        = string
   description = "Need to be unique within the account"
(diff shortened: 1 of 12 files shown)


platforms/azure/giantnetes/variables.tf
```diff
@@ -1,3 +1,14 @@
+variable "github_token" {
+  type        = string
+  description = "Your personal GITHUB token, used to get access to the private repository 'employees' to get the list of users."
```
calvix, line 3: how will this work for CI?

whites11: I guess we need to add the variable to conveyor jobs.
Will test that scenario before I merge.

calvix:
this repository was mostly considered public since it was meant to be publicly usable if anyone needs to, but lately, its got more and more focus towards our  platform that its almost impossible to sue outside of GS so I wonder if we should put that info somewhere or maybe even make the repo private

whites11:
> this repository was mostly considered public since it was meant to be publicly usable if anyone needs to, but lately, its got more and more focus towards our platform that its almost impossible to sue outside of GS so I wonder if we should put that info somewhere or maybe even make the repo private

Dropped a note in sig infra. Let's discuss this during next sync

platforms/azure/giantnetes/variables.tf
whites11: Addressed with https://github.com/giantswarm/conveyor/pull/87

whites11:
Manual tests worked very well on both AWS and Azure.
Trying to make the e2e tests green before I merge

whites11:
e2e tests are broken, but not because of my change.
I can't afford to work on those now, will drop a note in sig infra board and merge this branch.

PR #352, opened 2020-07-27
add node-exporter to bastions

glitchcrab:
towards giantswarm/giantswarm/issues/9390

tested on gauss/bastion2 - it is reachable from Prom.

### DIFF
--- a/templates/bastion.yaml.tmpl
+++ b/templates/bastion.yaml.tmpl
@@ -51,7 +51,7 @@ systemd:
         StartLimitInterval=300s
         StartLimitBurst=3
         TimeoutStartSec=0
-        Environment="IMAGE=quay.io/giantswarm/node-exporter:v0.18.0"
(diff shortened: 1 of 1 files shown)


templates/bastion.yaml.tmpl
```diff
@@ -30,6 +30,79 @@ systemd:
...
+        Description=node-exporter
+        Requires=docker.service
+        After=docker.service
+        [Service]
+        Restart=on-failure
+        RestartSec=1
+        StartLimitInterval=300s
+        StartLimitBurst=3
+        TimeoutStartSec=0
+        Environment="IMAGE=quay.io/giantswarm/node-exporter:v0.18.0"
```
calvix, line 54: can we go straight to 1.0.1 ?

glitchcrab: I've been testing 1.0.1 - i must've overwritten my changes whilst sorting some syntax issues i introduced. I've updated it to 1.0.1 :+1:

PR #364, opened 2020-08-05
Set default aws worker count to 4

corest:
Towards https://github.com/giantswarm/giantswarm/issues/12587


calvix:
unfortunately, this does not work for Adidas AWS CP as they have small CNI subnets

PR #365, opened 2020-08-05
Update flatcar linux version to 2512.2.1

corest:
Towards https://github.com/giantswarm/giantswarm/issues/11747


corest:
Tested on ginger. Works fine

PR #370, opened 2020-08-07
Allow scraping of TCs for TC prometheus

omegas27:
To allow our tenant cluster Prometheus to be able to scrape tenant clusters, we need to set the AWS_VPC_K8S_CNI_EXTERNALSNAT to false as otherwise as the POD CIDR of the tenant cluster prometheus server would not be allowed to talk with the tenant cluster. 
This was solved using hostNetwork: true for g8s-prometheus but this property is unusable with the prometheus operator as per https://github.com/prometheus-operator/prometheus-operator/issues/2630

### DIFF
--- a/templates/files/k8s-resource/aws-cni.yaml
+++ b/templates/files/k8s-resource/aws-cni.yaml
@@ -140,10 +140,6 @@ spec:
             - name: MINIMUM_IP_TARGET
               value: "2"
             ## Deviation from original manifest - 5
-            ## disable SNAT as we setup NATGW in the route tables
-            - name: AWS_VPC_K8S_CNI_EXTERNALSNAT
-              value: "true"
-            ## Deviation from original manifest - 6
             ## Explicit interface naming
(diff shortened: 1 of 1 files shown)


corest:
Let's not use references to private slack channels in public repos.
Please put a meaningful description for the issue in first message

templates/files/k8s-resource/aws-cni.yaml
```diff
@@ -140,10 +140,6 @@ spec:
             - name: MINIMUM_IP_TARGET
               value: "2"
             ## Deviation from original manifest - 5
-            ## disable SNAT as we setup NATGW in the route tables
-            - name: AWS_VPC_K8S_CNI_EXTERNALSNAT
-              value: "true"
```
TheoBrigitte, line 145: I would rather set the value to `false` explicitly here.

omegas27: Done

omegas27:
@giantswarm/chapter-sre  How can we ensure this gets deployed across all installations?

PR #380, opened 2020-08-27
Tag AWS route tables

wtty-fool:
Towards giantswarm/giantswarm#12940

### DIFF
--- a/modules/aws/vpc/vpc.tf
+++ b/modules/aws/vpc/vpc.tf
@@ -114,7 +114,8 @@ resource "aws_route_table" "cluster_vpc_private" {
   tags = merge(
     local.common_tags,
     map(
-      "Name", "${var.cluster_name}_private_${count.index}"
+      "Name", "${var.cluster_name}_private_${count.index}",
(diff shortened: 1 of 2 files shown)


corest (approved):

modules/aws/vpc/vpc.tf
```diff
@@ -114,7 +114,8 @@ resource "aws_route_table" "cluster_vpc_private" {
   tags = merge(
     local.common_tags,
     map(
-      "Name", "${var.cluster_name}_private_${count.index}"
+      "Name", "${var.cluster_name}_private_${count.index}",
+      "giantswarm.io/route-table-type", "private"
```
kopiczko, line 118: How about `giantswarm.io/cluster`?

wtty-fool: It's already there in `local.common_tags` https://github.com/giantswarm/giantnetes-terraform/blob/master/modules/aws/vpc/vpc.tf#L11

kopiczko (approved):

PR #381, opened 2020-09-02
Use explicit OIDC issuer url configuration

corest:
Towards https://github.com/giantswarm/giantswarm/issues/12939

### DIFF
--- a/platforms/azure/giantnetes/main.tf
+++ b/platforms/azure/giantnetes/main.tf
@@ -106,7 +106,7 @@ locals {
     "LogentriesToken"          = "${var.logentries_token}"
     "MasterCount"              = "${var.master_count}"
     "OIDCEnabled"              = "${var.oidc_enabled}"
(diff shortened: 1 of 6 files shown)


platforms/azure/giantnetes/main.tf
```diff
@@ -106,6 +106,7 @@ locals {
     "LogentriesToken"          = "${var.logentries_token}"
     "MasterCount"              = "${var.master_count}"
     "OIDCEnabled"              = "${var.oidc_enabled}"
+    "OIDCIssuerURL"            = "https://${var.oidc_issuer_dns}.${var.base_domain}"
```
kopiczko, line 109: Can't we hardcode `dex.g8s` here? What's the point of extracting `var.oidc_issuer_dns`?

kopiczko (approved):

platforms/azure/giantnetes/main.tf
corest: Don't like to have it here in case we change issuer

PR #383, opened 2020-09-15
rework-aws-cni-subnets1

calvix:
migrate AWS CNI subnets to CGNAT range to  use bigger ranges

steps to do after upgrading:
- recreate masters
- kill all pods

 after this PR is applied everywhere, the old AWS CNI resources will be deleted

applied to  gauss and ginger

towards https://github.com/giantswarm/giantswarm/issues/11899


templates/files/k8s-resource/calico-policy-only.yaml
```diff
@@ -505,7 +505,7 @@ spec:
             - name: FELIX_HEALTHENABLED
               value: "true"
             - name: FELIX_PROMETHEUSMETRICSENABLED
-            value: "true"  
+              value: "true"
```
calvix, line 508: fixed typo

xh3b4sd (approved):
Let's go when this works for you.

corest (approved):
lgtm

PR #384, opened 2020-09-15
Align aws cni manifest with upstream

calvix:
towards https://github.com/giantswarm/giantswarm/issues/11899

update aws CNI version to latest version and align  manifest with upstream ( and what we want to add into TC)

should fix race condition when aws cni and calico were trying to access same file


templates/files/k8s-resource/kube-proxy-ds.yaml
```diff
@@ -25,9 +25,7 @@ spec:
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
       tolerations:
-      - key: node-role.kubernetes.io/master
-        operator: Exists
-        effect: NoSchedule
+        - operator: Exists
```
calvix, line 28: not sure why this was an issue, but somehow kube-proxy did was not scheduled on not-ready nodes and it caused  deadlock since  aws-cni needed access to k8s api in order install CNI

calvix: so adding global ignore all taints

calvix:
tested on ginger and gauss

whites11 (approved):
Seems legit

njuettner (approved):
lgtm

PR #386, opened 2020-09-17
Use PartOf for etcd systemd unit

MarcelMue:
towards https://github.com/giantswarm/giantswarm/issues/12928


MarcelMue:
Closing here as this is not really helping the situation.

PR #387, opened 2020-09-29
Make docker pull optional before run

MarcelMue:
~~This is a somewhat controversial change:~~
~~Instead of pulling the image on every restart of the unit, we rely on `docker run` to pull it once and then use the local image.~~

~~Pro: We pull less often and it hopefully makes us a bit more tolerant to slow / flaky network,~~

~~Con: We have to use properly tagged images and ensure the correct image is actually on the machine.~~

Decided to make all docker pulls  not fail the unit itself.

Towards https://github.com/giantswarm/giantswarm/issues/12928

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
+### Fixed
 
 - Do not fail k8s-setup-network-env if docker pull fails.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Do not fail k8s-setup-network-env if docker pull fails.
+
 ## [1.4.0] - 2020-09-16
 
 ### Changed
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -750,9 +750,9 @@ systemd:
       Environment="NAME=%p.service"
       Environment="NETWORK_CONFIG_CONTAINER="
       ExecStartPre=/usr/bin/mkdir -p /opt/bin/
-      ExecStartPre=-/usr/bin/docker pull $IMAGE
       ExecStartPre=-/usr/bin/docker stop -t 10 $NAME
       ExecStartPre=-/usr/bin/docker rm -f $NAME
+      ExecStartPre=-/usr/bin/docker pull $IMAGE
       ExecStart=/usr/bin/docker run --rm --net=host -v /etc:/etc --name $NAME $IMAGE --verbose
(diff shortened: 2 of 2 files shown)


calvix:
>  it hopefully makes us a bit more tolerant to slow / flaky network,

the pull command has `-` at the start which means continue in case of failure so the unit would not fail if the pull fails

calvix (approved):

calvix:
if it anyhow helps sure

tfussell:
> the pull command has `-` at the start which means continue in case of failure so the unit would not fail if the pull fails

We didn't have that on `k8s-setup-network-env.service`, only on the other two. Not sure how the two approaches compare though.

kopiczko:
I think separate pull may give cleaner logs.

MarcelMue:
Just changed this around to just make the missing docker pull also optional / not fail the unit

templates/master.yaml.tmpl
```diff
@@ -750,7 +750,7 @@ systemd:
       Environment="NAME=%p.service"
       Environment="NETWORK_CONFIG_CONTAINER="
       ExecStartPre=/usr/bin/mkdir -p /opt/bin/
-      ExecStartPre=/usr/bin/docker pull $IMAGE
+      ExecStartPre=- /usr/bin/docker pull $IMAGE
```
tfussell, line 753:
```suggestion
      ExecStartPre=-/usr/bin/docker pull $IMAGE
```

templates/master.yaml.tmpl
```diff
@@ -750,7 +750,7 @@ systemd:
       Environment="NAME=%p.service"
       Environment="NETWORK_CONFIG_CONTAINER="
       ExecStartPre=/usr/bin/mkdir -p /opt/bin/
-      ExecStartPre=/usr/bin/docker pull $IMAGE
+      ExecStartPre=-/usr/bin/docker pull $IMAGE
       ExecStartPre=-/usr/bin/docker stop -t 10 $NAME
       ExecStartPre=-/usr/bin/docker rm -f $NAME
```
tfussell, line 755:
```suggestion
      ExecStartPre=-/usr/bin/docker stop -t 10 $NAME
      ExecStartPre=-/usr/bin/docker rm -f $NAME
      ExecStartPre=-/usr/bin/docker pull $IMAGE
```

tfussell: Not sure if the order matters, but just to be consistent.

MarcelMue: Fine with me :+1:

tfussell (approved):
LGTM. Not sure how much we can test it besides a simple check that it doesn't break anything.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
```
kopiczko, line 10:
```suggestion
### Fixed
```

kopiczko (approved):

PR #391, opened 2020-10-09
Change etcd disk size to 64gb

whites11:
This PR changes the etcd disk size for azure CPs to 64GBs.
On azure, the IOPS for disks are based on the disk size, and 64 is the smallest size we can use to have access to the following level of IOPS (240, currently 120).

This is important because we're hitting the IOPS limit and this is causing the master nodes to crash.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


corest (approved):
lgtm
Does this just resize etcd on the fly? Or how the migration is done?

whites11:
> lgtm
> Does this just resize etcd on the fly? Or how the migration is done?

I'm doing it manually

PR #400, opened 2020-10-28
Updating calico policy only deployment to apps/v1

njuettner:
calico-policy-only.yaml deployment needs to be updated to apps/v1

PM: https://github.com/giantswarm/giantswarm/issues/14006


corest (approved):
please upd changelog as well

njuettner:
Thanks 👍, updated

PR #404, opened 2020-11-02
Changed default VM size for azure master nodes to Standard_D4s_v3

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/13978

We decided to increate VM size for master nodes on azure because sometimes they can't keep the pace with the requested load.


glitchcrab (approved):
makes sense to me

PR #406, opened 2020-11-03
Reduce number of collectors used in node-exporter on bastions

corest:
Towards https://github.com/giantswarm/giantswarm/issues/14042

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Reduce number of collectors used in `node-exporter` on bastion node.
+
 ## [2.1.0] - 2020-11-02
 
 ### Added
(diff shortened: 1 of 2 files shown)


whites11 (approved):
Sounds good if works for you.

mazzy89:
Isn't easier to specify which collector we are enabling instead to specify which collector we are *enabling*? the mental process would tell me when I look at the PR or the deployed flags to go and check which are the other collectors. I agree with the direction but I see the node-exporter config as harder to read it in this specific case.

corest:
@mazzy89 config options are ordered - first go enabled, then disabled. What is hard to read exactly?

mazzy89:
So the only collector that we enable, they will be enabled. if we do not enable them explicitly no one of them will be enabled by default right? If this is true then I'm fine with that.

corest:
@mazzy89 this PR changes nothing regarding how node-exporter works. Just moving explicitly enabled to explicitly disabled. We do the same in tenant clusters

PR #424, opened 2020-12-09
Update azurerm to latest version 2.39

corest:

### DIFF
--- a/modules/azure/blob/blob.tf
+++ b/modules/azure/blob/blob.tf
@@ -1,16 +1,15 @@
 resource "azurerm_storage_account" "storage_acc" {
   name                      = "${var.cluster_name}config"
-  resource_group_name       = var.resource_group_name
   location                  = var.azure_location
   account_kind              = "BlobStorage"
   account_tier              = "Standard"
   account_replication_type  = "LRS"
   enable_https_traffic_only = true
+  resource_group_name   = var.resource_group_name
 }
 
 resource "azurerm_storage_container" "ignition" {
   name                  = "ignition"
-  resource_group_name   = var.resource_group_name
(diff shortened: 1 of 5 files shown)


modules/azure/blob/blob.tf
```diff
@@ -1,16 +1,15 @@
 resource "azurerm_storage_account" "storage_acc" {
   name                      = "${var.cluster_name}config"
-  resource_group_name       = var.resource_group_name
   location                  = var.azure_location
   account_kind              = "BlobStorage"
   account_tier              = "Standard"
   account_replication_type  = "LRS"
   enable_https_traffic_only = true
+  resource_group_name   = var.resource_group_name
```
whites11, line 8:
```suggestion
  resource_group_name       = var.resource_group_name
```

whites11 (approved):
LGTM when tested

corest:
Works locally, but fails in github action. Still need to debug

whites11:
FYI this was a blocker for bumping azurerm terraform in the past and apparently still is: https://github.com/terraform-providers/terraform-provider-azurerm/issues/2489#issuecomment-742027523

corest:
We don't have more subnets. And we need to switch to vmss and for that we need new provider version

sanderginn:
Hi @whites11 , I see you referenced my comment. I'd like to let you know that I had overlooked a resource and the proposed fix in the thread does seem to do the job.

whites11:
> Hi @whites11 , I see you referenced my comment. I'd like to let you know that I had overlooked a resource and the proposed fix in the thread does seem to do the job.

Thanks for letting me know @sanderginn

PR #427, opened 2020-12-10
Update terraform to 0.13.5 version

corest:
Towards https://github.com/giantswarm/giantswarm/issues/12286

### DIFF
--- a/modules/aws/vault/versions.tf
+++ b/modules/aws/vault/versions.tf
@@ -1,0 +1,11 @@
+terraform {
+  required_providers {
+    aws = {
+      source = "hashicorp/aws"
+    }
+    ignition = {
+      source = "terraform-providers/ignition"
+    }
+  }
+  required_version = ">= 0.13"
+}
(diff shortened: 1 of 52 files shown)


corest:
No dramatic changes.
Most of the updates are related to warning abut interpolated strings
Few changes regarding how 3rd party providers are managed now

whites11 (approved):

kopiczko (approved):

calvix (approved):

modules/aws/vault/versions.tf
```diff
@@ -0,0 +1,11 @@
+terraform {
+  required_providers {
+    aws = {
+      source = "hashicorp/aws"
```
mazzy89, line 4: small thing: below we pin providers, here no. is this wanted? should we pin providers everywhere?

mazzy89 (changes requested):
tiny question about pinning providers

mazzy89:
arrived too late 😇

PR #430, opened 2020-12-11
Set explicit longer timeouts on azure vm operations

corest:


whites11 (approved):
My autism dislikes the differen measurement unit (minutes vs hours) :sweat_smile: but ok :smile:

PR #433, opened 2020-12-14
Switch azure workers from availability sets to vmss

corest:


corest:
kudos to @whites11 for sorting this

modules/azure/vnet/lb-ingress.tf
```diff
@@ -93,3 +93,26 @@ resource "azurerm_lb_probe" "ingress_30011_lb" {
...
+
+resource "azurerm_lb_probe" "ssh" {
+  name                = "ssh-probe"
+  loadbalancer_id     = azurerm_lb.ingress_lb.id
+  resource_group_name = var.resource_group_name
+  protocol            = "tcp"
+  port                = 22
+  interval_in_seconds = 5
+  number_of_probes    = 2
+}
```
calvix, line 118: ssh lb probe ith ports 65000  and 22 semes like total cluster fuck to me,  I have no die what is that :p

can we add some comment that explain why it configured that way>

calvix (approved):
lgtm iwfy, do we need to adjust github actions for updating azure CPs?

modules/azure/vnet/lb-ingress.tf
corest: And it also fails during rollout 

```
Error: Creating/Updating Load Balancer "ghost-ingress-lb" (Resource Group "ghost"): network.LoadBalancersClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="CannotRemoveRuleUsedByProbeUsedByVMSS" Message="Load balancer rule /subscriptions/1be3b2e6-497b-45b9-915f-eb35cae23c6a/resourceGroups/ghost/providers/Microsoft.Network/loadBalancers/ghost-ingress-lb/loadBalancingRules/ingress-lb-fake-rule-for-node-health cannot be removed because the rule references the load balancer probe /subscriptions/1be3b2e6-497b-45b9-915f-eb35cae23c6a/resourceGroups/ghost/providers/Microsoft.Network/loadBalancers/ghost-ingress-lb/probes/ssh-probe that is used as health probe by VM scale set /subscriptions/1be3b2e6-497b-45b9-915f-eb35cae23c6a/resourceGroups/ghost/providers/Microsoft.Compute/virtualMachineScaleSets/ghost-workers. To remove this rule, please update VM scale set to remove the reference to the probe." Details=[]
```

whites11: So. The problem is how to rollout changes in VMSS instances.
Since we don't have one resource to be tainted for each worker, but only one VMSS object, I decided to rely on the `Rolling` upgrade mode feature.
That feature automatically reimages one node at a time (configurable how many in parallel) and waits for it to come back healthy before it goes on with other nodes.
To understand when a node is healthy, a load balancer probe has to be set. I decided to add a probe on port 22 (not ideal, but does the job for now).
Proble: azure requires the probe to be associated with a load balancer the VMSS belongs to and has to be referenced by an active forwarding rule.
Since I don't want to expose SSH through the load balancer, I used a random port, 65000.
Confusing? Hell yeah.
Alternatives? I don't think so.

whites11:
> lgtm iwfy, do we need to adjust github actions for updating azure CPs?

yes totally

modules/azure/vnet/lb-ingress.tf
calvix: Sure but let's put that as a comment above this :p SO someone reading it later can understand

corest: added comment

PR #436, opened 2020-12-15
Extend master/worker iam roles with required access for cluster-autoscaler

corest:

### DIFF
--- a/platforms/azure/giantnetes/variables.tf
+++ b/platforms/azure/giantnetes/variables.tf
@@ -76,7 +76,7 @@ variable "master_count" {
 variable "worker_count" {
   type        = string
   description = "Number of worker nodes to be created."
-  default     = "3"
+  default     = "4"
 }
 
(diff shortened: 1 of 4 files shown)


platforms/azure/giantnetes/variables.tf
```diff
@@ -76,7 +76,7 @@ variable "master_count" {
 variable "worker_count" {
   type        = string
   description = "Number of worker nodes to be created."
-  default     = "3"
+  default     = "4"
```
whites11, line 79: why did we change this? To comply with AWS?

whites11 (approved):

platforms/azure/giantnetes/variables.tf
corest: removed. I've used it for few installations where 3 nodes is not enough. But didn't want to merge as I hope autoscaler will solve this on its own

PR #439, opened 2020-12-16
Enable msi on azure masters/workers

corest:


whites11 (approved):
seems legit

PR #442, opened 2020-12-23
Retrieve vault token via IAM role access in AWS CPs

corest:

### DIFF
--- a/templates/master.yaml.tmpl
+++ b/templates/master.yaml.tmpl
@@ -796,7 +796,7 @@ systemd:
       ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/ssl/
       ExecStartPre=/usr/bin/mkdir -p /etc/tokens
       ExecStartPre=/bin/bash -c "while ! docker run --rm -v /etc/ssl/certs/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt {{.DockerRegistry}}/giantswarm/curl:7.67.0 -k -q --silent -o /dev/null https://{{ .VaultDomainName }};  do sleep 2s;echo wait for Vault;done;"
-      ExecStart=/bin/bash -c '\
+      ExecStart=/bin/bash -Oe -c '\
           export vault_token=$(\
           docker run --rm -i\
           -v /etc/ssl/certs/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt\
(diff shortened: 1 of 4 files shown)


templates/master.yaml.tmpl
```diff
@@ -805,7 +805,7 @@ systemd:
           -e VAULT_ADDR\
           -e CLUSTER_NAME\
           {{.DockerRegistry}}/giantswarm/vault:1.6.1\
-          login -method=aws role=$${CLUSTER_NAME}-master -format=json | jq -r .auth.client_token);\
+          login -method=aws role=$${CLUSTER_NAME}-master region={{ .AWSRegion }} -format=json | jq -r .auth.client_token);\
```
kopiczko, line 808: Looks ok but how was it working so far? Also may be a step backwards when it comes to multi-region support if we are planning to ever do that.

kopiczko (approved):

templates/master.yaml.tmpl
```diff
@@ -774,13 +776,49 @@ systemd:
...
+          export vault_token=$(\
+          docker run --rm -i\
+          -v /etc/ssl/certs/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt\
+          --net host\
+          --privileged=true\
+          -e VAULT_ADDR\
+          -e CLUSTER_NAME\
+          {{.DockerRegistry}}/giantswarm/vault:1.6.1\
+          login -method=aws role=$${CLUSTER_NAME}-master region={{ .AWSRegion }} -format=json | jq -r .auth.client_token);\
+          echo "VAULT_TOKEN=$${vault_token}" > /etc/tokens/node;\
```
calvix, line 809: when the aws command fails the   main  `ExecStart` command exits with 0 (due to the last command  doing just echo) thus silently falining

kopiczko: We can replace with `/bin/bash -Oe -c '\`

templates/master.yaml.tmpl
corest: It is a new service. We didn't have it before. Also multi-region support has nothing to do with this service.

kopiczko: Do you know what is this region used for? I quickly checked https://www.vaultproject.io/docs/auth/aws but it didn't make much sense to me. I thought IAM is global service.

corest: China uses a different STS endpoint, therefore, requires setting a corresponding region explicitly

corest: And there is example on the page you've referenced

```
The region used defaults to us-east-1, but you can specify a custom region like so:

$ vault login -method=aws region=us-west-2 role=dev-role-iam

When using a custom region, be sure the designated region corresponds to that of the STS endpoint you're using.
```

PR #446, opened 2021-01-19
Configure vault role with read permissions to VM/VMSS

corest:


whites11 (approved):
seems legit

PR #447, opened 2021-01-20
Add get vault token service for azure

corest:

### DIFF
--- a/templates/files/conf/get-vault-token
+++ b/templates/files/conf/get-vault-token
@@ -1,0 +1,25 @@
+#!/bin/sh
+
+set -uxo pipefail
+
+# Vault configuration
+export VAULT_ADDR=https://{{ .VaultDomainName }}
+export CLUSTER_NAME={{ .ClusterName }}
+
+# Azure metadata
+jwt="$(curl -s 'http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https://management.azure.com' -H Metadata:true | jq -r '.access_token')"
+subscription_id=$(curl -s -H Metadata:true "http://169.254.169.254/metadata/instance?api-version=2017-08-01" | jq -r '.compute | .subscriptionId')
+vm_name=$(curl -s -H Metadata:true "http://169.254.169.254/metadata/instance?api-version=2017-08-01" | jq -r '.compute | .name')
+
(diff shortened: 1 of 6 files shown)


templates/files/conf/get-vault-token
```diff
@@ -0,0 +1,25 @@
...
+set -uxo pipefail
+
+# Vault configuration
+export VAULT_ADDR=https://{{ .VaultDomainName }}
+export CLUSTER_NAME={{ .ClusterName }}
+
+# Azure metadata
+jwt="$(curl -s 'http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https://management.azure.com' -H Metadata:true | jq -r '.access_token')"
+subscription_id=$(curl -s -H Metadata:true "http://169.254.169.254/metadata/instance?api-version=2017-08-01" | jq -r '.compute | .subscriptionId')
+vm_name=$(curl -s -H Metadata:true "http://169.254.169.254/metadata/instance?api-version=2017-08-01" | jq -r '.compute | .name')
```
whites11, line 12: potentially VM name could have spaces in it, maybe it is better to put double quotes around this variable

```suggestion
vm_name="$(curl -s -H Metadata:true "http://169.254.169.254/metadata/instance?api-version=2017-08-01" | jq -r '.compute | .name')"
```

templates/master.yaml.tmpl
```diff
@@ -813,13 +811,36 @@ systemd:
       WantedBy=multi-user.target
   {{- end }}
 
+  {{ if eq .Provider "azure" -}}
```
whites11, line 814: I see you removed the `{{if eq .Provider "aws" }}` elsewhere, why do we still need this unit for azure only?

templates/files/conf/get-vault-token
corest: vm names are master/worker. So we don't have spaces there for sure. But just to be safe - sure

templates/master.yaml.tmpl
corest: aws unit us above this one. I just duplicated it to not mess with conditions in the unit body.

whites11: ok

whites11 (approved):
LGTM

PR #453, opened 2021-02-02
Allow node-exporter metrics on bastion from pod cidr

corest:
Towards https://github.com/giantswarm/giantswarm/issues/15441

### DIFF
--- a/platforms/aws/giantnetes/main.tf
+++ b/platforms/aws/giantnetes/main.tf
@@ -161,7 +161,7 @@ module "bastion" {
 
   arn_region             = var.arn_region
   aws_account            = var.aws_account
-  aws_cni_subnets     = var.aws_cni_subnets_v2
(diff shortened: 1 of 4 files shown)


platforms/aws/giantnetes/main.tf
```diff
@@ -161,6 +161,7 @@ module "bastion" {
 
   arn_region             = var.arn_region
   aws_account            = var.aws_account
+  aws_cni_subnets     = var.aws_cni_subnets_v2
```
calvix: `terraform fmt`  would be nice :trollface:

calvix (approved):

platforms/aws/giantnetes/main.tf
corest: download (11)

did that before your comment :)

calvix: yeah saw it after posting :p I bet it was my slow badwitdh

PR #458, opened 2021-02-17
Azure Managed Identity optional support

whites11:
This PR adds support to use azure managed identity to allow kubernetes components to login to azure APIs instead of a service principal.
If either the client ID or the client Secret of the service principal are not set, it will use the managed identity.

Please note this is a PR towards the 3.3 release branch


corest:
Why don't we merge this into master?

corest (approved):
lgtm if tested

whites11:
> Why don't we merge this into master?

I will backport to master as well, but for now I need a release with this change only

PR #466, opened 2021-03-31
Create bastions in VMware installations

glitchcrab:
This PR:
- adds support for a single bastion for an installation. the bastion resides in its own subnet which is attached to the same tier 1 router as the MC subnet so can access MC nodes.

### DIFF
--- a/modules/vmware/bastion/main.tf
+++ b/modules/vmware/bastion/main.tf
@@ -1,0 +1,67 @@
+resource "vsphere_folder" "main" {
+  path          = var.folder
+  type          = "vm"
+  datacenter_id = data.vsphere_datacenter.main.id
+}
+
+resource "vsphere_tag_category" "category" {
+  count = length(var.tags)
+
+  name        = var.tags[count.index]["scope"]
+  cardinality = "SINGLE"
+
+  associable_types = [
+    "VirtualMachine",
+    "Folder"
+  ]
+}
+
+resource "vsphere_tag" "tag" {
+  count = length(var.tags)
+
+  name        = var.tags[count.index]["tag"]
+  category_id = vsphere_tag_category.category[count.index].id
+}
+
+resource "vsphere_virtual_machine" "main" {
+  count = var.node_count
+
+  name             = format("%s-bastion-%s", var.cluster_name, count.index)
+  resource_pool_id = data.vsphere_compute_cluster.main.resource_pool_id
+  datastore_id     = data.vsphere_datastore.main.id
+  folder           = var.folder
+
+  num_cpus = var.cpus_count
+  memory   = var.memory
+  guest_id = data.vsphere_virtual_machine.main_template.guest_id
+
+  network_interface {
+    network_id = data.vsphere_network.bastion.id
+  }
+
+  disk {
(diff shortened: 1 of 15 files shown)


glitchcrab:
lets get this merged in - it works fine on the test installation.

modules/vmware/bastion/main.tf
```diff
@@ -0,0 +1,67 @@
...
+    size  = var.root_disk_size
+  }
+
+  clone {
+    template_uuid = data.vsphere_virtual_machine.main_template.id
+  }
+
+  extra_config = {
+    "guestinfo.ignition.config.data.encoding" = "base64"
+    "guestinfo.ignition.config.data"          = base64encode(var.ignition_data)
```
mazzy89, line 59: this kind force that the template is Flatcar and no other OS. is this expected?

mazzy89 (commented):
Overall looks fine. I have just a question about the direction of the OS for the bastion.

modules/vmware/bastion/main.tf
glitchcrab: for the moment yes, it made sense to stick with what we already use. unless there's any particular reason to use another OS it doesn't make sense to have more OSes to manage (at least in my eyes anyway).

it can always be changed in the future if necessary.

mazzy89: ok. for me at the current time is impossible to say for what this giantnetes-terraform will be used  in the VMware realm. TBH the master, vault and worker modules are even incomplete so this is why I raised this question. But yeah at the end of the day we need to make some choice here and stick with that.

platforms/vmware/giantnetes/main.tf
```diff
@@ -76,6 +84,59 @@ module "vault" {
   depends_on = [module.nsxt]
 }
 
+module "flatcar_linux" {
```
mazzy89, line 87: I'm a bit lost for what this is used. if you pass the template as variable to the module, then you are making the assumption that the template is already uploaded into VMware VSphere.

mazzy89 (changes requested):

platforms/vmware/giantnetes/main.tf
glitchcrab: good point - i meant to remove this (accidentally carried it over from another manifest)

glitchcrab: fixed.

platforms/vmware/giantnetes/variables.tf
```diff
@@ -150,6 +181,50 @@ variable "vsphere_folder" {
   description = "The path to the folder to put this virtual machine in, relative to the datacenter that the resource pool is in."
 }
 
+# Flatcar variables
+
+variable "flatcar_linux_channel" {
```
mazzy89, line 186: small change - not needed

mazzy89 (changes requested):
small change

platforms/vmware/giantnetes/variables.tf
glitchcrab: also sorted.

mazzy89 (approved):

PR #467, opened 2021-04-01
aws-increase-root-disk

calvix:
towards https://github.com/giantswarm/giantswarm/issues/16188

8GB disk has causes warning that we get `Warning: EBS Volumes are out of burst balance`, increasing the disk a little bit should give us better performance, its not critical or urgent but can improve overall speed of AWS CP

the cost increase should be low (around 10$ per month for whole CP)

### DIFF
--- a/modules/aws/master/variables.tf
+++ b/modules/aws/master/variables.tf
@@ -104,7 +104,7 @@ variable "volume_size_etcd" {
 
 variable "volume_size_root" {
   type    = string
-  default = 8
+  default = 30
 }
 
 variable "volume_docker" {
--- a/modules/aws/worker-asg/variables.tf
+++ b/modules/aws/worker-asg/variables.tf
@@ -79,7 +79,7 @@ variable "volume_size_etcd" {
 
 variable "volume_size_root" {
   type    = string
-  default = 8
+  default = 30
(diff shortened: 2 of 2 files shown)


modules/aws/master/variables.tf
```diff
@@ -104,7 +104,7 @@ variable "volume_size_etcd" {
 
 variable "volume_size_root" {
   type    = string
-  default = 8
+  default = 30
```
kopiczko, line 107: Won't it replace existing EBS volumes?

calvix: yes it will, but all the volumes except fro etcd are ephemeral and any  machine roll will create a new disk so its not an issue

kopiczko (approved):

whites11 (approved):

JosephSalisbury:
For my curiosity, was `30` chosen for a specific reason, or just semi-random?

calvix:
>  For my curiosity, was 30 chosen for a specific reason, or just semi-random?

not too big and not too small so you can say semi random

JosephSalisbury:
okay cool, i was curious if there was a threshold that 30 crossed or something. thanks!

PR #477, opened 2021-05-04
Update nginx-IC to 0.29.0

mig4:
To match the version deployed to on-prem MCs.

Towards: https://github.com/giantswarm/giantswarm/issues/12864

### DIFF
--- a/templates/files/k8s-resource/ingress-controller-deployment.yaml
+++ b/templates/files/k8s-resource/ingress-controller-deployment.yaml
@@ -67,7 +67,6 @@ spec:
         - --configmap=$(POD_NAMESPACE)/ingress-nginx
         - --annotations-prefix=nginx.ingress.kubernetes.io
         - --enable-ssl-chain-completion=false
-        - --enable-dynamic-certificates=true
         securityContext:
           allowPrivilegeEscalation: true
           capabilities:
--- a/templates/files/k8s-resource/ingress-controller-deployment.yaml
+++ b/templates/files/k8s-resource/ingress-controller-deployment.yaml
@@ -66,7 +66,6 @@ spec:
         - /nginx-ingress-controller
         - --configmap=$(POD_NAMESPACE)/ingress-nginx
         - --annotations-prefix=nginx.ingress.kubernetes.io
-        - --enable-ssl-passthrough
         - --enable-ssl-chain-completion=false
         - --enable-dynamic-certificates=true
         securityContext:
--- a/templates/files/k8s-resource/ingress-controller-deployment.yaml
+++ b/templates/files/k8s-resource/ingress-controller-deployment.yaml
@@ -56,7 +56,7 @@ spec:
       priorityClassName: system-cluster-critical
       containers:
       - name: nginx-ingress-controller
-        image: {{.DockerRegistry}}/giantswarm/nginx-ingress-controller:0.21.0
+        image: {{.DockerRegistry}}/giantswarm/nginx-ingress-controller:0.29.0
         args:
         - /nginx-ingress-controller
         - --default-backend-service=$(POD_NAMESPACE)/default-http-backend
--- a/templates/files/k8s-resource/ingress-controller-deployment.yaml
+++ b/templates/files/k8s-resource/ingress-controller-deployment.yaml
@@ -9,10 +9,15 @@ metadata:
     prometheus.io/port: '10254'
     prometheus.io/scrape: 'true'
 spec:
-  replicas: 3
   selector:
     matchLabels:
       k8s-app: nginx-ingress-controller
+  replicas: 3
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxSurge: 1
+      maxUnavailable: 2
   template:
     metadata:
       labels:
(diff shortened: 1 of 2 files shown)


calvix (approved):
is there a need for some change in configuration?

would be good to try run some CP with the new version before merge to is any potentional errors

mig4:
> is there a need for some change in configuration?
> 
> would be good to try run some CP with the new version before merge to is any potentional errors

Hi hi 😄 I was just writing a message to you in slack to ask for feedback on this 😄 We had a chat in standup and your name came up as the expert to check with.

So, OK, you're saying that I can pick an MC or two and just patch the `Deployment` on the cluster to see if this upgrade works as-is to make sure no nginx configuration are needed for compatibility with this version, right?

Yeah, I can do that. I'll report back when I have some data. Thanks!

JosephSalisbury:
also fyi @giantswarm/team-biscuit

mig4:
OK, tested this on _ghost_, I had to add a new API group to RBAC but after that it worked OK with no nginx config changes.

``` sh
❯ kubectl apply -f rbac-roles.yaml
clusterrole.rbac.authorization.k8s.io/system:coredns unchanged
clusterrole.rbac.authorization.k8s.io/nginx-ingress-controller configured                                                                                                                                                                
role.rbac.authorization.k8s.io/nginx-ingress-role unchanged
❯ kubectl set image deployment/nginx-ingress-controller nginx-ingress-controller=quay.io/giantswarm/nginx-ingress-controller:0.29.0
deployment.apps/nginx-ingress-controller image updated
```

I checked access to [Prometheus](https://prometheus.g8s.ghost.westeurope.azure.gigantic.io/ghost/), [AlertManager](https://alertmanager.g8s.ghost.westeurope.azure.gigantic.io/), [Grafana](https://grafana.g8s.ghost.westeurope.azure.gigantic.io/), etc. and they all work OK (I needed to clear auth cookie for AlertManager).

The [commit that upgraded nginx-IC to 0.29.0 in hive](https://github.com/giantswarm/hive/commit/2fe00f593f12bb8dc35d935bf1856a39abc9e971#diff-73f6e15cce1e938d4f46922b2994f5c2fb51cbe3fe3886805e6a555b57a3623b) also changes some `nginx-ingress-controller` parameters, notably:

- removes `--default-backend-service=$(POD_NAMESPACE)/default-http-backend` argument
- adds `--enable-dynamic-certificates=true` argument
- adds `securityContext` section

I don't know enough about these but if you think that's right I'm happy enough to make those changes here too to make these more aligned.

kopiczko:
> The [commit that upgraded nginx-IC to 0.29.0 in hive](https://github.com/giantswarm/hive/commit/2fe00f593f12bb8dc35d935bf1856a39abc9e971#diff-73f6e15cce1e938d4f46922b2994f5c2fb51cbe3fe3886805e6a555b57a3623b) also changes some `nginx-ingress-controller` parameters, notably:
> 
> * removes `--default-backend-service=$(POD_NAMESPACE)/default-http-backend` argument
> * adds `--enable-dynamic-certificates=true` argument
> * adds `securityContext` section
> 
> I don't know enough about these but if you think that's right I'm happy enough to make those changes here too to make these more aligned.

As I look at the PR it says it's aligned with nginx app which is what we have in WCs so I guess that makes sense. @calvix?

calvix:
Yeah aligning with both hive and nginx app would be great :)

mig4:
OK, updated to bring more in line with what's in hive and _nginx-ingress-controller-app_. There are two arguments used in _hive_ but not in _nginx-ingress-controller-app_ (AFAICS):

- `--enable-ssl-passthrough` - I dropped this as it breaks Prometheus, not 100% sure why but it doesn't appear to be used in _nginx-ingress-controller-app_ so I didn't investigate too deeply
- `--enable-dynamic-certificates=true` - with this nginx-IC complains - `Flag --enable-dynamic-certificates has been deprecated, Only dynamic mode is supported`, it also doesn't appear to be used in our app so I dropped it

Tested again with Prometheus, AlertManager, Grafana, Happa, etc. and all work OK.

JosephSalisbury:
cc @corest

mig4:
OK, if nobody vetoes I'm going to go ahead and merge this today :shipit:

mig4:
OK, merging :shipit: 

@calvix / @corest could you give me a bit of advice on getting this rolled out? Can I run it myself / does it require SRE with special permissions?

kopiczko:
You need to:

1. Release this repo.
2. Update version in installations like https://github.com/giantswarm/installations/pull/1727.
3. There is something about maintenance windows for adidas which is never going to be sorted. https://gigantic.slack.com/archives/C02EVLE9W/p1620047337147400 so probably don't care.. 
3. Run `opsctl update installation -i INSTALLATION -s all`.
4. Pray.
5. If it's Azure pray harder and retry.

kopiczko:
Oh yeah, I forgot because we are doing only cloud installations in Biscuit but if its KVM then you **definitely need to find upgrade window** because that rolls all the WC nodes. I have no idea how obtaining the window works cc @mazzy89 @glitchcrab.

mig4:
> Oh yeah, I forgot because we are doing only cloud installations in Biscuit but if its KVM then you **definitely need to find upgrade window** because that rolls all the WC nodes.

That's cool, I only need to do the cloud to match up with the version that's already on KVM. Thanks man, those are good instructions to get me going. I'll go grab some sacrifice for the Azure gods and get started :laughing:

PR #482, opened 2021-05-26
Use VMSS instead of VMs for master nodes on Azure.

whites11:


calvix:
you sure you want to have complex disk attaching in bash without any kind of error handling or retry logic?

we went for `go` for AWS attach dep disk tool

whites11:
> you sure you want to have complex disk attaching in bash without any kind of error handling or retry logic?
> 
> we went for `go` for AWS attach dep disk tool

It does retry. The unit is `Restart: on-failure` and it proved to be pretty effective (I made many MANY tests).
I see your point but the `go` route would me much longer and I spent to much time on this already.
Also consider MCs as we know them are going away soonish so this is going to be a temporary thing anyways.

Thanks a lot for your comment! <3

whites11:
This version has been running with zero problems on gremlin, godsmack, ghost, gollum and gauss for a couple of weeks.
I'm happy to make it an official release if you're ok with it

templates/files/conf/attach-etcd-disk
```diff
@@ -0,0 +1,152 @@
...
+  names="$(echo "$candidates" | jq -r .[].name)"
+  echo "Found some disks to try attaching: $names"
+
+  for diskname in $names
+  do
+    echo "Trying to attach disk $diskname to lun $LUN"
+    set +e
+    diskid="/subscriptions/${SUBSCRIPTION}/resourcegroups/${RESOURCEGROUP}/providers/Microsoft.Compute/disks/${diskname}"
+    # TODO un-hardcode registry
+    IMAGE="quay.io/giantswarm/azure-disk-attacher:latest"
```
fiunchinho, line 84: Can we configure the docker mirroring here? If not, maybe we can use ACR instead of quay here, since this is just for azure? https://portal.azure.com/#blade/Microsoft_Azure_ContainerRegistries/TagMetadataBlade/id/%2Fsubscriptions%2F1be3b2e6-497b-45b9-915f-eb35cae23c6a%2FresourceGroups%2Fgiantswarm-container-registry%2Fproviders%2FMicrosoft.ContainerRegistry%2Fregistries%2Fgiantswarm/repository/giantswarm%2Fazure-disk-attacher

whites11: good catch, let me check how we deal with images in MCs

whites11: According to biscuit we don't have registry mirroring in MCs, so this is not making nodes any less reliable than they already are. Does it work for you @fiunchinho ?

fiunchinho: I'm fine merging it as it is. But if we don't have mirroring, is there a reason NOT to use the azure registry, given that the image is already there?

whites11: I guess the reason is consistency to the other units that use quay.io?

fiunchinho: I'd argue reliability > consistency.

Anyway, your call.

fiunchinho (approved):

PR #484, opened 2021-06-17
Remove unused ami_owner variable

kopiczko:


calvix (approved):
if its really unused lets get rid if it

kopiczko:
We use `flatcar_ami_owner` now

PR #489, opened 2021-06-22
Add PE SSH keys to MC machines

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/17459


whites11 (approved):

whites11:
> Why did we only have SRE keys on MC machines? Security reasons?

This is just a backup. Normally we use vault signed keys to SSH (with opsctl SSH) but if that doesn't work we can ash directly using the keys

JosephSalisbury:
All good then :D

PR #493, opened 2021-07-09
Update azure storage udev rules

ericgraf:
NOTE* This is only need until https://github.com/kinvolk/init/pull/41 is included into flatcar image.

The `/boot` automount is failing on azure VMs. The automount is failing because there is no `.device` systemd unit for the disk. The root cause was discovered to be caused by udev rules. 

This fixes the udev rules to include TAG=systemd on azure storage devices. With this tag, systemd knows to create the `.device` unit.

 Related info:
> systemd will dynamically create device units for all kernel devices that are marked with the "systemd" udev tag
 https://www.freedesktop.org/software/systemd/man/systemd.device.html
 
 Related to work on https://github.com/giantswarm/giantswarm/issues/13754


ericgraf:
Tested working on `gremlin` and `ghost`. When deployed, the automount for `/boot` would function.

whites11 (approved):
LGTM

MarcelMue (approved):

MarcelMue:
Why is e2e failing though?

ericgraf:
> Why is e2e failing though?

I think they fail on all PRs. I clicked the unpause on the job.

PR #496, opened 2021-07-15
Update AZ number in the AWS installation guide

kopiczko:
Now there are 3 AZs in CN


marians:
As there are two regions in China, is this correct for both?

kopiczko:
> As there are two regions in China, is this correct for both?

Yes. But still I'd suggest checking this every time.

marians (approved):

PR #504, opened 2021-08-13
Add kubelet flag to better detect memory eviction threshold

ericgraf:
This change is already pushed to WCs. 

Flag description:
> kernelMemcgNotification, if set, the kubelet will integrate with the kernel memcg notification to determine if memory eviction thresholds are crossed rather than polling. Dynamic Kubelet Config (beta): If dynamically updating this field, consider that it may impact the way Kubelet interacts with the kernel. Default: false

https://v1-20.docs.kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/


stone-z (approved):
LGTM. For context, done for WCs here: https://github.com/giantswarm/k8scloudconfig/pull/922

PR #508, opened 2021-08-20
Remove AWS `vpn_instance` module

kopiczko:


platforms/aws/giantnetes/variables.tf
```diff
@@ -301,12 +301,6 @@ variable "external_ipsec_subnet_1" {
   default     = "172.18.0.4/30"
 }
 
-### VPN with custom instance 
-variable "vpn_instance_enabled" {
```
kopiczko, line 305: Just in case I grepped installations and didn't find it being set anywhere.

corest:
We don't have anyone requesting this now. But if there would be someone with request to have China installation, but not having direct connect - we would still need this.

corest (approved):

kopiczko:
That's legit. I dropped a card in our inbox to discuss that. But I'm not sure we want to have CN installation without DC.

kopiczko:
We talked and don't want a new CN installation without DC connection set. Otherwise connections are really unreliable.

PR #509, opened 2021-08-23
aws-fix-hostnetworkpods-limit-worker

calvix:
towards https://github.com/giantswarm/giantswarm/issues/18186

file was copied from TC and it looks like we have one less  hostnetwork pod on worker on MC


anvddriesch (approved):
lgtm. Maybe check if this repo has a changelog that needs updating.

njuettner (approved):

PR #529, opened 2021-12-14
Correctly calculate max pods for control plane nodes

AverageMarcus:
Related to https://github.com/giantswarm/giantswarm/issues/20118

The `setup-kubelet-environment` script is never run with any arguments so the previous check for if it's a master node of not would always end up thinking it was a worker node.
This change creates different script content based on the node type at templating stage.


whites11 (approved):
tested, worked like a charm. Well done!

PR #530, opened 2021-12-15
bump flatcar and kubernetes

whites11:


whites11:
just a version bump, going with a ping

PR #532, opened 2021-12-15
Revert flatcar version

whites11:


kopiczko:
What went wrong?

whites11:
> What went wrong?

- falco app does not work.
- there is a problem with node port services using `externalTrafficPolicy: Local` (this breaks nginx ingress controller and thus the whole cluster)

PR #544, opened 2022-01-18
In azure, force service principal credentials to be explicitely set.

whites11:
On azure, when running terraform from the SREs machines, we used to rely on the locally set up credentials to interact with azure API.

This proved to be unreliable and ineffective for a couple of reasons:

- lighthouse access that we request to customers for our own access does not have all required permissions
- CI/CD uses a different authentication method, leading to problems

With this PR we force the presence of service principal credentials as variables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


T-Kukawka (approved):

fiunchinho (approved):
LGTMIIWFY

njuettner (approved):

PR #550, opened 2022-01-26
Remove hardcoded coredns

whites11:
This PR deletes the hardcoded coredns manifests from the terraform repo and uses the managed app instead


templates/files/conf/k8s-addons
```diff
@@ -128,9 +128,17 @@ $KUBECTL -n kube-system delete deploy calico-kube-controllers --ignore-not-found
...
 $KUBECTL -n kube-system delete daemonset nginx-ingress-controller --ignore-not-found=true
 
+# delete coredns manifests
+$KUBECTL -n kube-system delete svc coredns --ignore-not-found=true
+$KUBECTL -n kube-system delete serviceaccount coredns --ignore-not-found=true
+$KUBECTL -n kube-system delete cm coredns --ignore-not-found=true
+$KUBECTL -n kube-system delete deploy coredns --ignore-not-found=true
+$KUBECTL delete clusterrolebinding system:coredns --ignore-not-found=true
+$KUBECTL delete clusterrole system:coredns --ignore-not-found=true
+
```
calvix, line 138: what happens if this is rerun after the app is installed?

calvix: maybe some if that will check if the app exists and delete only if it does not exists yet?

templates/files/k8s-resource/psp-bindings.yaml
```diff
@@ -11,12 +11,6 @@ subjects:
   name: calico-kube-controllers
   namespace: kube-system
 {{- end }}
-- kind: ServiceAccount
-  name: coredns
-  namespace: kube-system
-- kind: ServiceAccount
-  name: nginx-ingress-controller
-  namespace: kube-system
```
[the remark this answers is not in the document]
whites11: accident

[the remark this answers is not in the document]
whites11: restored

templates/files/conf/k8s-addons
whites11: very good point. I thought of this while doing nginx ingress controller (PR coming), didn't check for coredns.
Checking

whites11: how about this?

calvix: looks good if it works as expected :p

calvix (approved):

PR #551, opened 2022-01-26
Use managed app to deploy nginx ingress controller (AWS).

whites11:
This PR removes hardcoded nginx ingress controller manifests and uses latest managed app instead.
One side effect of this is that nginx ingress controller pods are going to potentially be scheduled on master nodes as well.
This needs a change in the load balancer, coming in a different PR.


templates/files/conf/k8s-addons
```diff
@@ -128,6 +127,23 @@ $KUBECTL -n kube-system delete deploy calico-kube-controllers --ignore-not-found
 # delete legacy nginx daemonset
 $KUBECTL -n kube-system delete daemonset nginx-ingress-controller --ignore-not-found=true
 
+# delete nginx ingress controller installed by terraform <= 5.10
+$KUBECTL -n kube-system delete deploy default-http-backend --ignore-not-found=true
```
whites11, line 131: checked resource names, none of them overlap with the managed app.
If you still think it's insecure we might add a check as we did with the coredns app

templates/files/conf/k8s-addons
```diff
@@ -128,6 +127,23 @@ $KUBECTL -n kube-system delete deploy calico-kube-controllers --ignore-not-found
...
+$KUBECTL -n kube-system delete svc default-http-backend --ignore-not-found=true
+$KUBECTL -n kube-system delete cm ingress-nginx --ignore-not-found=true
+$KUBECTL -n kube-system delete serviceaccount nginx-ingress-controller --ignore-not-found=true
+$KUBECTL -n kube-system delete deploy nginx-ingress-controller --ignore-not-found=true
+$KUBECTL -n kube-system delete cm ingress-nginx --ignore-not-found=true
+$KUBECTL -n kube-system delete svc nginx-ingress-controller --ignore-not-found=true
+$KUBECTL delete clusterrolebinding nginx-ingress-controller --ignore-not-found=true
+$KUBECTL delete clusterrole nginx-ingress-controller --ignore-not-found=true
+$KUBECTL -n kube-system delete rolebinding nginx-ingress-controller --ignore-not-found=true
+$KUBECTL -n kube-system delete role nginx-ingress-role --ignore-not-found=true
```
calvix, line 141: the naming is different from the app?

calvix: otherwise same comment as with coredns

whites11: yeah see my other comment. There is no overlap, but nothing ensures it'll never happen.
The goal should be to remove those delete commands from a future release, as soon as all MCs will be upgraded to this release.
Open to hear your opinions about

calvix (approved):

PR #552, opened 2022-01-26
Move Ingress ELB resources in own module (AWS).

whites11:
Since we switched to the nginx ingress controller managed app, the nginx ingress controller pods may end up in any worker or master node (it was workers only before).
Since we use `externalTrafficPolicy: local`, only nodes running an nginx ingress controller pod will respond to the ELB.
That means we need to add the master nodes to the ingress ELB (it is managed by terraform).

This makes it necessary to create the ELB before master nodes.
This PR extracts the ingress ELB creation to its own module so it can be run before masters and worker nodes, and changes master nodes in order to add themselves to the ELB.

WARNING: this is a breaking change and will require manual intervention to be applied. I will take care of doing so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Move Ingress ELB resources in own module (AWS).
 - Use managed app to deploy nginx ingress controller (AWS).
 - Use managed app to deploy coredns (AWS).
 - Bump flatcar to 3033.2.0 (AWS).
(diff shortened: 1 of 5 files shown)


calvix:
wait a minute, the nginx app creates a custom  service type loadblaner which will create a brand new ingress

whites11:
> wait a minute, the nginx app creates a custom service type loadblaner which will create a brand new ingress

no it does not thanks to this https://github.com/giantswarm/giantnetes-terraform/blob/master/templates/files/k8s-resource/ingress-controller-app.yaml#L14

whites11:
> how does that work together with DNS? as on WC we have external DNS which updates the DNS for the ingress

we decided to stick with the terraform managed load balanced to avoid the need of external-dns

calvix:
oh alrighty

calvix (approved):

PR #554, opened 2022-01-28
Switch Azure from using calico for CNI to Calico with user provided routes

whites11:


templates/master.yaml.tmpl
```diff
@@ -38,34 +38,30 @@ storage:
...
         id: 0
       group:
         id: 0
       contents:
-        source: "data:text/plain;charset=utf-8;base64,{{ index .Files "k8s-resource/calico-crd-installer-rbac.yaml" }}"
-{{ else }}
+        source: "data:text/plain;charset=utf-8;base64,{{ index .Files "k8s-resource/calico-policy-only.yaml" }}"
+{{ end }}
+{{ if ne .Provider "aws" }}
     - path: /srv/calico-all.yaml
```
whites11, line 65: Still need this manifest on azure because we need to cleanup old resources before we switch to policy-only

fiunchinho (approved):

PR #555, opened 2022-01-28
Switch Azure Load Balancer to be Standard rather than Basic.

whites11:
Since we need to recrete the load balancer as part of having nginx installed as managed app, I took the chance to change the load balancer to be standard instead of basic.

This will require manual work to be done, I'll take care of that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Switch Azure Load Balancer to be Standard rather than Basic.
 - Switch Azure from using calico for CNI to Calico with user provided routes.
 - Move Ingress ELB resources in own module (AWS).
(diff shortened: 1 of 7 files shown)


calvix (approved):
so much love for azure

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Switch Azure Load Balancer to be Standard rather than Basic.
 - Switch Azure from using calico for CNI to Calico with user provided routes.
 - Move Ingress ELB resources in own module (AWS).
 - Use managed app to deploy nginx ingress controller (AWS and Azure).
```
njuettner, line 15: LGTMIIWFY.

I think the indent is a bit ahead vs line 16 and below.

njuettner (approved):

CHANGELOG.md
njuettner: nevermind I saw that wrong 🤦🏻

PR #558, opened 2022-01-31
Fix CIDR in Azure CNI masquerading unit.

whites11:
routing was broken for pods towards vault host because of a wrong CIDR in the masquerading unit for azure cni


whites11:
ping ping

PR #560, opened 2022-02-01
Remove migration commands from k8s-addons (was needed to upgrade from < 6 to 6.0.0).

whites11:
removed code that was just needed for migrating


fiunchinho (approved):

calvix (approved):
we should ensure that the version with migration code is not skipped

PR #566, opened 2022-02-07
corends-use-custom-configmap

calvix:


templates/files/k8s-resource/coredns-app.yaml
```diff
@@ -1,3 +1,14 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: coredns-user-values
+  namespace: giantswarm
+data:
+  values: |
+    cluster:
+      kubernetes:
+        clusterDomain: {{ .ClusterDomain }}
```
calvix, line 10: we override what ever is in app catalog here

whites11 (approved):
smart

PR #570, opened 2022-02-10
Bump nginx ingress controller app to 2.9.0 and enable `enable-ssl-chain-completion` flag.

whites11:
Needed by viking because they use custom SSL certs that need this feature.
It used to be enabled in old nginx ingress controller, was disabled when we moved to the managed app.


fiunchinho (approved):
LGTMIIWFY

PR #572, opened 2022-02-11
Azure termination events

whites11:
This PR adds self-drain feature on termination for azure worker nodes to make upgrade smoother


modules/azure/worker-as/worker-as.tf
```diff
@@ -1,20 +1,29 @@
-resource "azurerm_virtual_machine_scale_set" "workers" {
+resource "azurerm_linux_virtual_machine_scale_set" "workers" {
```
whites11, line 1: azurerm_virtual_machine_scale_set is deprecated, azurerm_linux_virtual_machine_scale_set is the newer alternative and I need it for enabling termination events

PR #575, opened 2022-02-17
Azure master roll

whites11:
Towards: https://github.com/giantswarm/installations/runs/5234134673?check_suite_focus=true

- Using the new `azurerm_linux_virtual_machine_scale_set` terraform resource in place of the deprecated `azurerm_virtual_machine_scale_set`.
- Enable termination notification on masters to make azure-scheduled-events effective.


whites11:
going with a ping

PR #578, opened 2022-02-23
Split worker nodes in different ASGs, each one having a single AZ

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/20953

this PR splits the worker nodes into different ASGs, in order for cluster-autoscaler to be able to scale the right AZ based on the needed availability zone.

This is the first step, where the old ASG is still present. It has to be removed in a follow up PR in order to avoid downtimes during rollout.

It goes along with this change in the deployment CI https://github.com/giantswarm/installations/pull/1980

currently running in ginger.

### DIFF
--- a/modules/aws/worker-asg/worker-asg.tf
+++ b/modules/aws/worker-asg/worker-asg.tf
@@ -28,7 +28,7 @@ resource "aws_cloudformation_stack" "worker_asg_single_az" {
         "DesiredCapacity": "2",
         "HealthCheckType": "EC2",
         "HealthCheckGracePeriod": 300,
-        "LaunchConfigurationName": "${aws_launch_configuration.worker.name}",
+        "LaunchConfigurationName": "${aws_launch_configuration.worker_asg_single_az.name}",
         "LoadBalancerNames": [
           "${var.cluster_name}-worker"
         ],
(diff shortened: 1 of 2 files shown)


modules/aws/worker-asg/worker-asg.tf
```diff
@@ -89,6 +89,81 @@ resource "aws_cloudformation_stack" "worker_asg" {
...
+            "Key": "k8s.io/cluster-autoscaler/${var.cluster_name}",
+            "Value": "true",
+            "PropagateAtLaunch": false
+          }
+        ],
+        "VPCZoneIdentifier": [${jsonencode(element(var.worker_subnet_ids, count.index))}]
+      },
+      "UpdatePolicy": {
+        "AutoScalingRollingUpdate": {
+          "MinInstancesInService": "${var.worker_count}",
```
calvix, line 148: should this be 1?

calvix:
does it mean 6 workers mean 6 ASGs split across 3 AZs?

calvix (approved):
LGTM IWFY

modules/aws/worker-asg/worker-asg.tf
whites11: This will go away complete in the next release so it doesn't really matter

whites11:
> does it mean 6 workers mean 6 ASGs split across 3 AZs?

Nope, it's one for each subnet we create, should be 3 for all MCs

calvix:
okok

PR #588, opened 2022-03-15
Switch to out-of-tree controller manager for Azure

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/20965

This PR disables in-tree azure cloud controller manager features and installs the out-of-tree alternatives.
As a consequence it also switches to the CSI volume provisioners.

### DIFF
--- a/templates/files/apps/azure/azuredisk-csi-driver.yaml
+++ b/templates/files/apps/azure/azuredisk-csi-driver.yaml
@@ -30,5 +30,5 @@ spec:
     secret:
       name: ""
       namespace: ""
-  version: v1.16.0-gs1
+  version: v1.16.0-gs2
 
--- a/templates/files/apps/azure/azuredisk-csi-driver.yaml
+++ b/templates/files/apps/azure/azuredisk-csi-driver.yaml
@@ -30,5 +30,5 @@ spec:
     secret:
       name: ""
       namespace: ""
-  version: v1.13.0-gs1
+  version: v1.13.0-gs2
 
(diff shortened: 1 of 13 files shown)


templates/files/conf/k8s-addons
```diff
@@ -136,7 +136,7 @@ MANIFESTS="default-storage-class.yaml
 for manifest in $MANIFESTS
 do
     while
-        $KUBECTL apply -f /srv/$manifest
+        $KUBECTL apply --force -f /srv/$manifest
```
whites11, line 139: Needed by the default storage class change

templates/files/apps/azure/azuredisk-csi-driver.yaml
```diff
@@ -0,0 +1,34 @@
...
+  namespace: kube-system
+  namespaceConfig: {}
+  userConfig:
+    configMap:
+      name: ""
+      namespace: ""
+    secret:
+      name: ""
+      namespace: ""
+  version: v1.13.0-gs1
```
fiunchinho, line 33: How come we need custom builds for all these components?

fiunchinho (approved):

templates/files/apps/azure/azuredisk-csi-driver.yaml
whites11: it's not custom builds, I'm using the same versioningstrategy as we do for cluster-autoscaler.
I'm using the upstream version, but with `-gs<n>` suffix to allow us to version our helm charts.
Does it make sense?

PR #589, opened 2022-03-18
Set `fs.inotify.max_user_instances` to 1024 to avoid `Too many open f…

whites11:
…iles` error.


calvix (approved):

calvix:
what is the default value? 256?

whites11:
> what is the default value? 256?

128

PR #593, opened 2022-03-29
Bump azurerm provider to 3.0.2.

whites11:
Towards: https://github.com/giantswarm/roadmap/issues/966

We need to bump to latest version of azurerm terraform provider, but that requires a few changes here and there


whites11:
will move on with a ping as I am blocked by this

PR #596, opened 2022-03-30
Allow setting custom tags to all resources in Azure.

whites11:
This PR allows setting custom tags for all resources.
It's needed by ista and talanx so far


fiunchinho (approved):
I don't know enough about terraform to really approve this but LGTMIIWFY

PR #600, opened 2022-04-08
allow-capa-s3-buckets

calvix:


whites11 (approved):

modules/aws/vpc/vpc.tf
```diff
@@ -50,6 +50,13 @@ EOF
       "Effect": "Allow",
       "Resource": "arn:${var.arn_region}:s3:::etcd-backups.giantswarm.io/*"
     },
+    {
+      "Sid": "CAPA",
```
QuentinBisson, line 54: I don't want to be a pain in the butt but do you think you could rename it to CAPA-Rules to match the pattern of the other policies?

PR #616, opened 2022-06-01
Cilium

whites11:
Towards: https://github.com/giantswarm/roadmap/issues/1136

This switches from calico and kube-proxy to cilium.
Been running on all test MCs and GS' prod ones for a few weeks without big problems.


njuettner (approved):
lgtm

PR #622, opened 2022-06-23
Disable k8s API fairness

whites11:


MarcelMue (approved):

calvix:
just out of curiosity why do we need it?

MarcelMue:
> just out of curiosity why do we need it?

We are trying it out for the vodafone incident - https://github.com/giantswarm/giantnetes-terraform/pull/622#issuecomment-1164169373

PR #647, opened 2022-09-13
Change permission of etcd key file in order to make `etcd-kubernetes-…

whites11:
…resources-count-exporter` app to run with unprivileged user.

Towards: https://github.com/giantswarm/giantswarm/issues/23643


glitchcrab (approved):
seems reasonable

PR #656, opened 2022-09-29
Allow bucket access for auditing

QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/23045#issuecomment-1262244695

This PR allows customers and us to send data to buckets containing `-giantswarm-`.

cc @paurosello what do you think about this?

### DIFF
--- a/modules/aws/vpc/vpc.tf
+++ b/modules/aws/vpc/vpc.tf
@@ -69,7 +69,7 @@ EOF
       "Principal": "*",
       "Action": "*",
       "Effect": "Allow",
-      "Resource": "arn:${var.arn_region}:s3:::*-giantswarm-*/*"
+      "Resource": "arn:${var.arn_region}:s3:::*-giantswarm-audit-logs*/*"
     }
   ]
 }
--- a/modules/aws/vpc/vpc.tf
+++ b/modules/aws/vpc/vpc.tf
@@ -51,7 +51,7 @@ EOF
       "Resource": "arn:${var.arn_region}:s3:::etcd-backups.giantswarm.io/*"
     },
     {
-      "Sid": "CAPA-Rule",
+      "Sid": "CAPA",
       "Principal": "*",
       "Action": "*",
       "Effect": "Allow",
--- a/modules/aws/vpc/vpc.tf
+++ b/modules/aws/vpc/vpc.tf
@@ -51,7 +51,7 @@ EOF
       "Resource": "arn:${var.arn_region}:s3:::etcd-backups.giantswarm.io/*"
     },
     {
-      "Sid": "CAPA",
+      "Sid": "CAPA-Rule",
       "Principal": "*",
       "Action": "*",
       "Effect": "Allow",
@@ -63,6 +63,13 @@ EOF
       "Action": "*",
       "Effect": "Allow",
       "Resource": "arn:${var.arn_region}:s3:::*-g8s-*"
+    },
+    {
+      "Sid": "Giantswarm-Audit-Rule",
+      "Principal": "*",
+      "Action": "*",
+      "Effect": "Allow",
+      "Resource": "arn:${var.arn_region}:s3:::*-giantswarm-*/*"
     }
   ]
 }


mcbenjemaa:
I can't review this PR. Maybe Pair reviewing helps.

modules/aws/vpc/vpc.tf
```diff
@@ -63,6 +63,13 @@ EOF
       "Action": "*",
       "Effect": "Allow",
       "Resource": "arn:${var.arn_region}:s3:::*-g8s-*"
+    },
+    {
+      "Sid": "Giantswarm-Audit-Rule",
+      "Principal": "*",
+      "Action": "*",
+      "Effect": "Allow",
+      "Resource": "arn:${var.arn_region}:s3:::*-giantswarm-*/*"
```
QuentinBisson, line 72: Adds the /* at the end so objects can be updated (important for auditing)

modules/aws/vpc/vpc.tf
```diff
@@ -51,7 +51,7 @@ EOF
       "Resource": "arn:${var.arn_region}:s3:::etcd-backups.giantswarm.io/*"
     },
     {
-      "Sid": "CAPA",
+      "Sid": "CAPA-Rule",
```
TheoBrigitte, line 54: Why are we changing this ? I prefer to avoid unnecessary changes.

modules/aws/vpc/vpc.tf
TheoBrigitte: Are we sure the bucket name format will always follow this `*-giantswarm-*/*` ?

TheoBrigitte (changes requested):

modules/aws/vpc/vpc.tf
hervenicol: I think it's the aim of the PR. If there is no "-giantswarm-" in the bucket name, access is forbidden.

QuentinBisson: Yes, Vodafone uses account-id-giantswarm-audit-logs and I think it makes sensé that we en force that in the future

modules/aws/vpc/vpc.tf
QuentinBisson: I'll revert it on monday. It just makes non sense that this rule does not end with -Rule when all the others do

modules/aws/vpc/vpc.tf
QuentinBisson: @stone-z I thing I did not express the issue properly, I;m sorry.
What this changes does is to not block access to s3 buckets containing giantswarm from our management clusters at the VPC level ([s3 gateway endpoint](https://docs.aws.amazon.com/vpc/latest/privatelink/vpc-endpoints-s3.html) that ensure network access goes through the internal amazon network and not the internet) but there are still iam roles that prevents any action on buckets that is not authorized. The role for fluent-bit at customers is limited to s3 calls to this bucket anyway :)

But I agree we can enforce that the buckets should contain giantswarm-audit-logs.

I would like in a near future to have this bucket using crossplane so we provide audit logs by default but that's a far away dream

modules/aws/vpc/vpc.tf
QuentinBisson: Done

QuentinBisson:
@TheoBrigitte can you check this again?

TheoBrigitte (approved):
Works for me

QuentinBisson:
@whites11 can you run this on a test installation and viking?

whites11:
you can do it yourself on a test installation but not viking until its tested.
Add `RELEASE_VERSION=master` here https://github.com/giantswarm/installations/blob/master/gauss/terraform/bootstrap.sh#L15
push and run `opsctl update installation -i gauss -s all`.

For next time, please test it before merging.

QuentinBisson:
Thanks Chris I will

QuentinBisson:
Tested successfully on Gauss
image

@whites11 can I ask for you help in running it?

whites11:
Ok so I assume your test was successful.
Now you can make a new giantnetes-terraform minor release and upgrade Gauss to that release (copy from another installation for the Syntax).
Then opsctl update installation again. 
You can do the same with Viking if you want to

whites11:
Otherwise it'll be deployed with next round of upgrades at the end of this month

QuentinBisson:
Thanks ;) I'll run viking then

PR #673, opened 2022-11-02
[AWS] Fix IAM permission for master nodes when using cross-account KMS

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/24307


whites11:
Tested on flamingo, made prometheus to work!

njuettner (approved):

PR #679, opened 2022-12-11
Add missing labels for scheduler and controller manager

QuentinBisson:
Working on installing the prometheus agent on Vintage MCs, I noticed that the scheduler and controller manager were missing labels set on default service monitors https://github.com/giantswarm/prometheus-meta-operator/pull/1094

This PR fixes that.

At the same time, I noticed that the scheduler healthcheck is actually checking for the life of the controller manager (using port 10257). I'm guessing both scheduler and controller-manager are running in the host network, which allowed this behavior.

This image is a call to port 10257 using kubectl port-forward on the k8s scheduler (see leader_election_status metric at the end)
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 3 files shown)


marieroque (approved):

QuantumEnigmaa (approved):

QuentinBisson:
Tested on ginger

PR #688, opened 2023-02-02
Add loki S3 bucket and permission

paurosello:
Enable loki monitoring with an S3 bucket and permission on the node instance profile

Towards https://github.com/giantswarm/roadmap/issues/1849

### DIFF
--- a/modules/aws/master/master-iam.tf
+++ b/modules/aws/master/master-iam.tf
@@ -76,6 +76,28 @@ resource "aws_iam_role_policy" "master" {
     {
         "Effect": "Allow",
         "Action": [
+            "s3:ListBucket",
+            "s3:PutObject",
+            "s3:GetObject",
+            "s3:DeleteObject" 
+        ],
+        "Resource": [
+            "arn:${var.arn_region}:s3:::${var.aws_account}-${var.cluster_name}-g8s-loki",
+            "arn:${var.arn_region}:s3:::${var.aws_account}-${var.cluster_name}-g8s-loki/*"
+        ]
+    },
+    {
+        "Effect": "Allow",
+        "Action": [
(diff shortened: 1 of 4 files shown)


modules/aws/master/master-iam.tf
```diff
@@ -73,6 +73,28 @@ resource "aws_iam_role_policy" "master" {
...
+            "s3:DeleteObject" 
+        ],
+        "Resource": [
+            "arn:${var.arn_region}:s3:::${var.aws_account}-${var.cluster_name}-g8s-loki",
+            "arn:${var.arn_region}:s3:::${var.aws_account}-${var.cluster_name}-g8s-loki/*"
+        ]
+    },
+    {
+        "Effect": "Allow",
+        "Action": [
```
whites11, line 91: why no resource filter here?

paurosello: this is about S3 access points, not s3 buckets itself. Currently we are not using any Access Points in our VPCs so I cannot filter by anything.

Its mentioned by loki that these permissions are needed, it might fail if we don't add it.

whites11 (approved):

paurosello:
Enabling terrafom in https://github.com/giantswarm/installations/pull/2112/files

QuantumEnigmaa (approved):
Yeah !

QuentinBisson:
Hey @paurosello do you mind doing the same for mimir so we can get a head start?

hervenicol:
> Hey @paurosello do you mind doing the same for mimir so we can get a head start?

We said this would be done manually on our designated test installations first, then when we want to push it to all installations would be pushed either at next MC updates (once per month), or even earlier if it's urgent.

QuentinBisson:
Alright

PR #713, opened 2023-06-09
Remove CSIMigration flags

njuettner:
Issue: https://github.com/giantswarm/roadmap/issues/2536
For reference: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates-removed/


whites11 (approved):
me love that

PR #722, opened 2023-07-10
Enable auditd

QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/24992

As seen here https://www.flatcar.org/docs/latest/setup/security/audit/#enabling-auditd it looks like we enabled the auditd rules but not the audit daemon which is actually needed to have audit logs in a file


whites11 (approved):

QuentinBisson:
image
Works like a charm

QuentinBisson:
@whites11 can I release and roll this on viking?

PR #735, opened 2023-10-25
Add tempo buckets

QuentinBisson:
This PR adds tempo and mimir bucket on vintage installations. This is needed so we can try out tempo and migrate to mimir in a bit


QuentinBisson:
Tested on Gaia

whites11 (approved):

PR #742, opened 2023-11-29
nginx-ingress-controller-app: Increase `controller.resources.requests.memory` to 500Mi.

Gacko:
Towards https://github.com/giantswarm/giantswarm/issues/28509.


Gacko:
For reviewers: I not only reordered the manifest and removed default / empty values, but also changed the `catalog` to `control-plane`. Therefore I ensured the app got pushed and is still getting pushed to this catalog beforehand.

Additionally I increased the memory requests since on some MCs Ingress NGINX Controller was scaling up due to too few memory while consuming only little CPU and therefore requesting too much CPU in total.

ubergesundheit (approved):

PR #752, opened 2024-04-30
bump apps and disable PSPs

whites11:
Latest 1.24 release with disabled PSPs


templates/files/apps/common/cilium-app.yaml
```diff
@@ -96,4 +99,4 @@ spec:
     secret:
       name: ""
       namespace: ""
-  version: 0.12.0
+  version: 0.21.0
```
AndiDog, line 102: Is it intentional that we upgrade Cilium (v1.13.x to v1.14.x)?

whites11: yeah

AndiDog (approved):

PR #770, opened 2024-09-30
Update ingress-nginx on vintage MCs

ubergesundheit:
Update the ingress-nginx container image tag to use v1.11.2 on vintage MCs

Towards https://github.com/giantswarm/adidas/issues/1298

The change in here already has been applied to some aws vintage management clusters


mcharriere (approved):

mcharriere:
changelog

